PR #143, opened 2015-10-28
#MIFOSAC-81 fixed

chhavip:
With reference to https://mifosforge.jira.com/browse/MIFOSAC-81, I have displayed loan officer and client groups but to display number of loan cycles, I would require more information as to how it is obtained from the data obtained.


ishan1604:
Needs a rebase.

PR #148, opened 2015-11-06
General project update

fomenkoo:
Project update: 

- [General] butterknife updated to 7.0.1
- [General] support-v4 updated to 23.0.1
- [General] Google Play Services updated to 8.3.0 (latest)
- [General] ActionBarActivity(deprecated) replaced with AppCompatActivity
- [General] compileSdkVersion set to 23
- [General] added 'org.apache.http.legacy' to android section of build.gradle (apacheHttp was deprecated in api level 22)

Material design & improvements: 

- [Arhitectural] Added BaseActivity class (which will contain basic for all activities methods),
- [Arhitectural] Added android databinding support (https://developer.android.com/tools/data-binding/guide.html), that allows us to do the sings similar to butterknife, but more (MVVM)
- [Arhitectural] Added SpManager class, to manage changes related to shared preferences (in cached way, because when storing in shared prefs changes are not immidiate)
- [Arhitectural] Created styles files for each type of view, that allows to make xml layouts more readable and compact.
- [UI] Login activity re-design
- [Validation] Validation refactoring
- [Build] moved buildScript from module's build.gradle to project root (it should be there)
- [Build] updated gradle plugin to 1.3.1


fomenkoo:
Here is how LoginActivity looks now with material design: 

https://www.youtube.com/watch?v=-p9mIfJTzb0&feature=youtu.be

:-)

PR #151, opened 2015-11-08
Material update

fomenkoo:
Material design & improvements:

[Arhitectural] Added BaseActivity class (which will contain basic for all activities methods),
[Arhitectural] Added android databinding support (https://developer.android.com/tools/data-binding/guide.html), that allows us to do the sings similar to butterknife, but more (MVVM)
[Arhitectural] Added SpManager class, to manage changes related to shared preferences (in cached way, because when storing in shared prefs changes are not immidiate)
[Arhitectural] Created styles files for each type of view, that allows to make xml layouts more readable and compact.
[UI] Login activity re-design
[Validation] Validation refactoring
[Build] moved buildScript from module's build.gradle to project root (it should be there)
[Build] updated gradle plugin to 1.3.1


fomenkoo:
@ishan1604 hi, thanks! 
1. I've split all the commits into 2 parts - first one is related to general project update, second is about using material design. If you still want me to split commits under second pull request - just tell me :) 
2. Minimal support level is 11 (which means android 3.0 and above). If you lover it to 10 (android 2.3) it will work. 
3. Probably you're right, but I've seen several situations in my project, when setting was set synchronously, and it wasn't updated to the time it was need again. CachedValue doesn't uses any kind of reflection or something, it just 'synchronized' block which allows us to be sure that setting was set. 

But I can remove it if you want :)

PR #152, opened 2015-12-09
Revert "Material update"

ishan1604:
Reverts openMF/android-client#151


bnasim:
@ishan1604 
Why did you revert Material update, after you merged it?

PR #153, opened 2015-12-21
Material design library

fomenkoo:
- Added com.android.support:design:23.0.1 - support library (for material design)<br/>
[Google Material Design support library] (http://developer.android.com/tools/support-library/features.html#design) <br/>

- Design library doesn't compiles with old google-play-services library, so, only the way to use it - update google-play-services. I've used com.google.android.gms:play-services-location:8.3.0 - it ships as separate package, and this way the size of APK will be much more lower. I've updated receiving location as it described in official documentation. <br/> [Google Play Services - receiving location updates](http://developer.android.com/training/location/retrieve-current.html)<br/>

It works correctly, and without this changes it's not possible to work with official way to implement material design.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,11 @@ language: android
 jdk: oraclejdk7
 android:
   components:
-    - build-tools-20.0.0
     - extra-android-support
+    - extra-google-google_play_services
     - extra-android-m2repository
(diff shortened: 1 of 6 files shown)


fomenkoo:
Thanks a lot Ishan!

PR #158, opened 2015-12-28
Fixed logout, url reinitialized

chhavip:
Earlier the instanceUrl was saved to the entire https://demo.openmf.org/mifosng-provider/api/v1 after a login which wasn't reset to demo.openmf.org on logout which was leading to repeated url and failure to login again without uninstall. Fixed by reinitializing the variable on logout.


chhavip:
@ishan1604 , please review the PR and let me know if changes are needed.

PR #161, opened 2016-01-20
Survey First Two Screens

bnasim:


mifosng-android/src/main/java/com/mifos/mifosxdroid/adapters/ClientChooseAdapter.java
```diff
@@ -0,0 +1,55 @@
...
+            holder = (ViewHolder) convertView.getTag();
+        }
+        item = getItem(position);
+        holder.clientnameView.setText(item.getDisplayName());
+        TextDrawable drawable = mDrawableBuilder.build(String.valueOf(item.getDisplayName().charAt(0)), mColorGenerator.getColor(item.getDisplayName()));
+        holder.clientimageView.setImageDrawable(drawable);
+        return convertView;
+    }
+
+    class ViewHolder {
```
ishan1604: This class should be static. 

If you want to know the reasons check this links : 

http://stackoverflow.com/questions/31302341/what-difference-between-static-and-non-static-viewholder-in-recyclerview-adapter

http://stackoverflow.com/questions/13516835/why-in-viewholder-pattern-should-the-viewholder-class-be-static

http://stackoverflow.com/questions/12368302/is-it-crucial-for-performance-to-have-viewholder-as-static-in-a-viewholder-patte

mifosng-android/src/main/java/com/mifos/mifosxdroid/adapters/ClientChooseAdapter.java
```diff
@@ -0,0 +1,55 @@
...
+        private ImageView clientimageView;
```
ishan1604: It would be better to use this convention.

iv_clientImage
tv_clientName

So that when accessing the object the user knows what is the type of the View (Whether TextView or ImageView)

mifosng-android/src/main/java/com/mifos/mifosxdroid/adapters/SurveyListAdapter.java
```diff
@@ -0,0 +1,78 @@
...
+        final Survey survey = listSurvey.get(i);
+
+        viewHolder.tv_survey_name.setText(survey.getName());
+        viewHolder.tv_description.setText(survey.getDescription());
+        return view;
+    }
+
+    public static class ViewHolder{
+        @InjectView(R.id.tv_survey_name)
+        TextView tv_survey_name;
```
ishan1604: This Looks Awesome :+1:

bnasim:
So Ishan, do you want me to submit another pull request or you can make minor change.

ishan1604:
Please resubmit

PR #168, opened 2016-02-01
Material survey

bnasim:
Following functionalities included in PR
Fetch active surveys from server.
Included client list to choose a client from
Parse the Survey JSON to get question and options
support multiple options
Capture user responses
Submit Scorecard

I wrote the procedure to submit Scorecard but currently doesn't work because still there is problem in Survey API as mentioned by Markus in developer mailing list at
https://groups.google.com/forum/#!topic/mifosdeveloper/M7E29NYxpVA


ishan1604:
Needs Squashing

PR #174, opened 2016-02-22
Payment API + Pref Manager + Squash + Comments

fomenkoo:
Hi, here is the first part of offline mode implementation. 
**1. API calls architecture.**

**ApiManager** is managing all out-coming API calls. In such a way, we can keep track with internet connection, and unsaved entities. In case your device offline - entity will be saved to database, and then synced with server side (to be implemented). 

**BaseApiManager** is responsible to create API services (they now are in separated interface classes). 
Authentication service is flexible, and with ApiEndpoint class it's now easy to change server URL to connect with (that's not a good solution to re-create whole API class in LoginActivity). 

Consider writing your new API calls in corresponding interface classes (all of them are located in **_com.mifos.api.services_** package)

**2. Preferences management** 
In old architecture we had something like this: 
```
    private SharedPreferences sharedPreferences;
    sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
    SharedPreferences.Editor editor = sharedPreferences.edit();
    editor.putString(Constants.INSTANCE_URL_KEY, instanceURL);
    editor.apply(); 
```
That was almost in each fragment and activity. Refactoring this, I was able to remove about 200 lines of duplicate code. 

I've created new class **PrefManager**. Using it is very simple: 
```
    PrefManager.putString(PREF_KEY_PROMISE,"I'll not copy-paste code anymore");
```
or you can define pref code in manager, and use it like this:
```
    PrefManager.setToken("");
    PrefManager.getToken();
```

**3. Old Holo theme files** 
I've removed all old resource files, and themes that are dedicated to old theme. With this, apk size was reduced by 2-3 megabytes. 

**4. Refactoring**
I've refactored most of the code.
Consider not using global static variables, especially for context. 

P.S: I've squashed commits & changed comments


ishan1604:
@fomenkoo There are some files missing LICENSE Headers. Please add them.

fomenkoo:
@ishan1604 can you please review all the code once instead of making me re-merging this?

ishan1604:
Wait for a couple of days then, if you need complete review of the code.

PR #177, opened 2016-02-29
MIFOSAC-113 Important validations added, unimportant removed

vjs3:
Issues resolved in this PR are:
1.There were many unnecessary validation fields has been added to the create group, create client and create centers functionality, as I discussed with mentors compulsory fields should be same in both web as well as android interface, hence a solution involving it has been provided.
2.There was no validation for group's name (it is compulsory for a group to have some name, webapp contains it) which has been added too.

P.S-Merge conflict is resolved!

### DIFF
--- a/mifosng-android/src/main/java/com/mifos/mifosxdroid/online/CreateNewCenterFragment.java
+++ b/mifosng-android/src/main/java/com/mifos/mifosxdroid/online/CreateNewCenterFragment.java
@@ -5,8 +5,10 @@
 
 package com.mifos.mifosxdroid.online;
 
+import android.app.Activity;
 import android.os.Bundle;
 import android.support.v4.app.DialogFragment;
+import android.text.TextUtils;
 import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.View;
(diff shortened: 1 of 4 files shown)


vjs3:
@ishan1604 this is a separate issue with no overlapping, please merge it if you find it correct.

vjs3:
@ishan1604 as even after creating a separate branch for separate issue, commits are coming together.So please help me out with this situation else I will not be able to commit other bugs till the time this PR don't get merged.Please help me so that I can keep working as I am stuck right now.

ishan1604:
@vjs3 I don't think you should remove any validation that is already there. Anything that prevents failure is good!

* Make sure you don't send in any Commented out code!
* Also don't override methods until you need them. Like in this case `onAttach()` this is what is unimportant.

vjs3:
@ishan1604 if we have web and android client for same product isn't it good to have both have same checks?I asked it in community as well and they said the similar thing that's why I only commented them out so that they can be used in future.Also there are some checks which I have added, can I send a new PR containing the missing checks only?

nellyk:
@vjs3 thank you for raising this issue i have been working on them and i added the center creation ,group creation, and added more fields to the client creation. I'll have a look at them cc: @ishan1604

vjs3:
@nellyk please let me know if I can help in solving some bugs in any way.
Thanks!

PR #181, opened 2016-03-03
GIS Integration

fomenkoo:
Hi! :) 

This PR contains integration with Google Maps and location features:

**- Pinpointing client location**

**MifosMapActivity** - activity that holding SupportMapFragment (to make it work on API level < 16). 
**PinpointClientActivity** - responsible for onClick callbacks, pinpointing, and saving client location. 

To store field officer path I've created **LocationService**. It designed to work with location in 'Optmial Power Mode'. To request saving path we should connect to it through ServiceConnection, and to call appropriate method. 

During recording I'm showing now a notification with recording status. 

Data is stored as list of LatLng coordinates, and before sending it to backend it will be converted to PolyLine - https://developers.google.com/maps/documentation/utilities/polylinealgorithm
This will allow us to keep cross-platform way to store and display route. 
Saving is not implemented yet, because I can't find appropriate methods in api.

### DIFF
--- a/mifosng-android/build.gradle
+++ b/mifosng-android/build.gradle
@@ -112,6 +112,8 @@ dependencies {
     // You must install or update the Support Repository through the SDK manager to use this dependency.
     compile 'com.android.support:appcompat-v7:23.1.1'
     compile 'com.google.android.gms:play-services-location:8.3.0'
+    compile 'com.google.maps.android:android-maps-utils:0.4.2'
+
     compile 'com.android.support:design:23.1.1'
     // You must install or update the Support Repository through the SDK manager to use this dependency.
     compile 'com.android.support:support-v4:23.0.1'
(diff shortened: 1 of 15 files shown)


ishan1604:
@fomenkoo Great!

Does it start capturing location when the Destination is added by Field Officer or it keeps on capturing the data irrespective of the fact where the field officer is going.?

fomenkoo:
Now it's not starting capturing automatically, I think we can discuss it this evening :)

I've checked API documentation as well as source code, and I haven't found any things related to location. 

Event old code in ClientDetailsFragment (that saves GPS data to datatable is not working, because of backend answering with 'wrong permission'). 

I wanted to discuss this things at our meeting today.

ishan1604:
Sure let's discuss it in the meeting today. :)

PR #182, opened 2016-03-03
Fixed NumberFormatException caused by OfflineCenterInputActivity

Sushant-M:
When a user entered characters instead of integers, the app used to crash, added an NumberFormatException to handle such an incident.


Sushant-M:
@ishan1604 Please look into it, thanks :)

ishan1604:
1. Remove the `Log.d` method call.
2. Do not use hardcoded strings. Store them in Strings.xml and refer it from there.

Sushant-M:
I have made the changes, I don't want to create a new pull request and as far as I know it won't update the commit if the request is closed, so could you please open it again so that it updates? :+1:

PR #183, opened 2016-03-05
MIFOAC-117 Showing the Connection Status resolved

mayank-kgp:
On Login Page If network connection is not available then snackbar would show message about network connection.

screenshot_2016-03-05-18-11-29


mayank-kgp:
@edcable   @ishan1604  
Please look into this.

PR #184, opened 2016-03-06
MIFOSX-118 Showing Progress Bar in opening Client list resolved

mayank-kgp:
After clicking on client list screen appears blank for some time. Sometimes it goes upto 15 second. It would be better to show progress bar at this stage.
Now it shows a progress bar .

Before -  
screenshot_2016-03-06-14-52-31

Now -- 

screenshot_2016-03-06-16-16-04


tarun0:
The referenced pull request is resolving a different issue than the one resolved here.

Issue #185, opened 2016-03-08
App crashes on selecting documents from toolbar.


Sreedev:
- Select a client from the list of clients.
     - Select the documents from the toolbar.
     - App crashes. (Resources$NotFoundException exception)

    03-07 23:24:45.419    2185-2185/? E/AndroidRuntime﹕ FATAL EXCEPTION: main Process: com.mifos.mifosxdroid, PID: 2185 android.content.res.Resources$NotFoundException: String resource ID #0xe2c0
            at android.content.res.Resources.getText(Resources.java:312)
            at android.widget.TextView.setText(TextView.java:4422)
            at com.mifos.mifosxdroid.adapters.DocumentListAdapter.getView(DocumentListAdapter.java:84)
            at android.widget.AbsListView.obtainView(AbsListView.java:2346)
            at android.widget.ListView.measureHeightOfChildren(ListView.java:1281)
            at android.widget.ListView.onMeasure(ListView.java:1188)
            at android.view.View.measure(View.java:18794)
            at android.widget.RelativeLayout.measureChildHorizontal(RelativeLayout.java:715)
            at android.widget.RelativeLayout.onMeasure(RelativeLayout.java:461)
            at android.view.View.measure(View.java:18794)
            at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:5951)
            at android.widget.FrameLayout.onMeasure(FrameLayout.java:194)
            at android.view.View.measure(View.java:18794)
            at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:5951)
            at android.widget.LinearLayout.measureChildBeforeLayout(LinearLayout.java:1465)
            at android.widget.LinearLayout.measureVertical(LinearLayout.java:748)

Sreedev:
I am working on this.

prempalsingh:
I have already fixed this in #195 and it has been merged. This can be closed.

therajanmaurya:
All Error resolved during uploading Documents in #335 and #336.
@ishan1604 Please close this issue.

PR #189, opened 2016-03-09
Addded a try catch and resolved crash on documents from client list.

Sreedev:
Resolved Issue #185.


ishan1604:
I'd like you fix the code that raises the exception and not just catch it.

Sreedev:
Thats great ishan.

PR #211, opened 2016-03-11
Upload Client Image from Gallery

### LINKED ISSUE #197: Upload an Image from gallery to update client's profile picture.
It would be good to have an upload function on the image instead of using camera in the client's page

therajanmaurya:
Users can change their profile picture by choosing an image from the gallery. Earlier it was only possible by taking a picture from the camera.

Fixes #197


therajanmaurya:
Already implemented.

PR #213, opened 2016-03-12
MIFOSAC-81 | Added Group name, Loan Cycle and Staff name to the Client General Page

tarun0:


mifosng-android/src/main/java/com/mifos/mifosxdroid/online/ClientDetailsFragment.java
```diff
@@ -375,6 +389,10 @@ public void inflateClientsAccounts() {
         App.apiManager.getClientAccounts(clientId, new Callback<ClientAccounts>() {
             @Override
             public void success(final ClientAccounts clientAccounts, Response response) {
+                if(!clientAccounts.isNullOrEmpty())
+                    tv_loan_cycle_number.setText(String.format("%d", clientAccounts.getLoanAccounts().get(0).getLoanCycle()));
```
tarun0: Need clarification on this part. Since every loan is associated with a cycle number, so I've doubt as to which one should be shown on the Client Page. I've emailed on the mailing list as well and am waiting for the reply. The cycle number of the very first loan are being shown till then.

PR #221, opened 2016-03-13
fix MIFOSAC-119

prempalsingh:
fixes MIFOSAC-119. Changing date of birth doesn't affect date of active client and vice-versa any more.


ishan1604:
I'd rather introduce an `id` field in the `MFDatePicker` class that can uniquely identify the MFDatePicker instance. Since you've duplicated code, which could have been avoided using this way. 

```
public void onDatePicked(String date, int datePickerId) {

   if (datePickerID == mfDatePicker.getId()) {
      tv_submissionDate.setText(date);
   } else if (datePickerID == newDatePicker.getId()) {
      tv_dateofbirth.setText(date);
   }        
}
```

PR #230, opened 2016-03-17
MIFOSAC -123 Activity goes blank if there is no question in survey resolved

mayank-kgp:
If there is no question in any survey activity goes blank and user is confused whether the questions are uploading or what?
So there should be a toast giving information about empty survey.

screenshot_2016-03-17-22-06-20


ishan1604:
Hello Mayank

Good thinking about the User Experience. 

It would be great if you could add this message to Strings.xml so that it can be localised with other strings. 

> It is not recommended to hard code user facing strings in an android application

 :smiley:  

I'll merge it as soon as it is fixed!

mayank-kgp:
@ishan1604  ya, I am fixing it very soon. 
I will inform you after fixing

mayank-kgp:
@ishan1604  I have added in String.xml 
But there is a problem. This activity is removed from mater branch and renamed as fragment. 
So it would be better if I close this pull request and open a new as it would not take much time.

PR #234, opened 2016-03-19
fix: :construction_worker: Make foolproof against NumberFormatException.

### LINKED ISSUE #233: Offline Mode - App crash in case user doesn't enter a valid Branch or Staff ID Integer
In offline mode app crashes due to NumberFormatException in case the user doesn't enter a valid Integer for Branch or Staff ID.

RowlandOti:
Prevent app crash in case user doesn't enter a valid Integer - NumberFormatException. This will prevent the app from crashing whenever either Branch Id or Staff Id are not valid Integers

fix #233
Resolves: #233


RowlandOti:
@ishan1604 kindly review this!

PR #223, opened 2016-03-14
ImageLoadingAsyncTask to Picasso, for image loading in different thread.

therajanmaurya:
PR #207  squashed . @ishan1604 Please review and merge.

PR #235, opened 2016-03-20
Loadmore ClientList and Refresh the ClientList fix.

therajanmaurya:
There was not Refresh functionality properly enable and LoadMore was not properly working.

Now it is fixed and officers can Refresh the Client List and view the Next Pagination Client List using scrolling down the list. 

@fomenkoo Please Review 
Thanks.

### DIFF
--- a/mifosng-android/src/main/java/com/mifos/mifosxdroid/online/ClientListFragment.java
+++ b/mifosng-android/src/main/java/com/mifos/mifosxdroid/online/ClientListFragment.java
@@ -46,10 +46,11 @@ public class ClientListFragment extends MifosBaseFragment {
 
     private View rootView;
     private List<Client> clientList = new ArrayList<>();
-    private int offset = 0;
+    private boolean areMoreClientsAvailable = true;
+    private int totalFilteredRecords = 0;
+    private boolean shouldCheckForMoreClients = false;
+    private boolean loadmore = false;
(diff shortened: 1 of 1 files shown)


fomenkoo:
Hi Rajan,

Good implementation - nothing to say more :)
Thanks to contributing! :)

@ishan1604 can you close this request?

therajanmaurya:
Please Also look #223 

Thanks.

ishan1604:
@therajanmaurya  :+1: for the implementation.

Here is my review of your code.

I think we can stick to naming the variables like this 
```
private boolean clientAvailable = true;
```

and not like this

```
private boolean ClientAvailable = true;
```

I also feel it would be great to call this variable `areMoreClientsAvailable` 

And since `CHECK_MORE_CLIENT` is not a constant it doesn't have to be named this way. 

only `static final` variables must be named like this. 

and since we are not writing our code in C we can try to refrain ourselves from using `1` & `0` Flags and use `boolean` instead. 

so you can modify the variable to ``` boolean shouldCheckForMoreClients```

Also please **squash** your commits when submitting PRs for a particular feature or bug fix.

therajanmaurya:
Thanks @ishan1604,

I have changed variable names and squash. it's look good.
Please Review.

ishan1604:
Looks good! Thanks

PR #239, opened 2016-03-26
submission and birthdate became independent

ahmed-fathy-aly:
fixes MOFISAC-124
when creating a new client and changing the submission date, the birth date doesn't change
And added an activity test for it.


ishan1604:
Thanks for the PR

PR #244, opened 2016-04-01
add navigation drawer

ahmed-fathy-aly:
MIFOSAC-136 adds a navigation drawer to the dashboard activity
modifies activity tests to work with these changes and adds a few more
![Screenshot]
(http://i.imgur.com/rePno8k.png)


ishan1604:
1. The Profile Name should be of the signed in user and not of any client.
2. Logout option should be offered by clicking on the drop down and not on the Navigation Drawer.

ahmed-fathy-aly:
@ishan1604  Thanks for your feedback
1- The profile name is of the signed in user indeed, The loadClientDetails() method in MifosBaseActivity makes an API call to fetch the logged in client's details, shows the display name and accountNumber in the account header then loads the profile picture
2- Sure, I'll move Logout to the options menu

therajanmaurya:
@ahmed-fathy-aly good implementation.
You are using an external library "'com.mikepenz:materialdrawer:4.5.1'" for navigation drawer and android-client  already have Design Support Library in build.gradle, using design library, you can easily implement Material Navigation Drawer that will support all Android 2.1 or higher devices.

So Please, can you do that with design support library.
Here are  some references you can use.
[http://developer.android.com/reference/android/support/design/widget/NavigationView.html](url)
[http://android-developers.blogspot.in/2015/05/android-design-support-library.html](url)
[https://github.com/chrisbanes/cheesesquare](url)

Note: Please use project libraries first.

ahmed-fathy-aly:
@therajanmaurya  But that library looks good, follows material design guidelines, compatible down to API 10(the app's min API level is , and its integration requires much less change than using the design support library.
That's why I think that's library is a much better choice
Lets hear what @ishan1604  thinks about that

therajanmaurya:
@ahmed-fathy-aly, First : there is no guarantee that this library will always compatible with coming API  versions like Android N and more coming API. it can be depreciated in future. 
So, we should use design support.

Second : if we are adding libraries for our simplicity then I hope, we will cross  65000 method policy soon.
and we have to set up MultiDex in the project that make APK size bigger.

ahmed-fathy-aly:
The library is well maintained, so I don't think we should worry about incompatibility with upcoming API version. Indeed, I'm more worried about us reinventing the wheel and writing the code ourselves which makes us need to worry about maintaining it in the future and instead of having the library maintainers take care of any compatibility issues in the future, we need to take care of them ourselves.
Let's wait for ishan's comment, I want to hear his opinion

ahmed-fathy-aly:
@ishan1604 
*updated*
moved log out from the navigation drawer to the options menu
also added a "Groups" item to the navigation drawer, Thanks @nellyk  for pointing out there was a fragment which lists all groups implemented.

ishan1604:
I'd agree with @therajanmaurya we don't need to use this external library for no reason.

+1 for the reasons @therajanmaurya 

@ahmed-fathy-aly Please implement the Navigation drawer without this library. It might require a little effort. But trust me, you'll feel good about it when the PR is merged :smiley:

ahmed-fathy-aly:
@ishan1604  Ok, I'll work on it :)

ahmed-fathy-aly:
*updated*
@ishan1604  Now it uses android's design support library instead of the external one
and updated the related activity tests accordingly
I edited the description to have the latest screenshot

ishan1604:
One last thing, can you shift implementation to Glide and remove ion?
And then we are good to merge your PR

ahmed-fathy-aly:
*updated*
@ishan1604  Done, Now It uses Glide instead of Ion.

xiprox:
I feel like a few little bits are slightly off here and there. I presume things have changed a bit after switching to the Design Support Library though. Can you provide a screenshot of how this looks right now?

ahmed-fathy-aly:
@xiprox I updated the screenshot after switching to the design support library
So the screenshot in the description is how it looks now
Can you clarify which bits are off ?

xiprox:
Okay, so:
**1.** How about we get the status bar translucent and have the navigation drawer extend behind it? You should be able to do this by setting the theme's status bar color to transparent and then setting the actual color to the DrawerLayout.
**2.** Is that a border around the image/avatar? We usually don't have borders in navigation drawer avatars. A slight (1-2dp?) shadow/elevation could be fine though.
**3.** What's that extra space under the name in the header for? How about we get rid of that? 16dp padding (just like the sides) should be enough, I'd say.
**4.** Where do you get the icons from? They must be from some other collection as they are inconsistent with the general look of material icons. Some lines in the icons for Survey and Path Tracker are too thin for a material icon. The spacing between the human silhouettes seems too little as well. How about you choose something from [here](https://design.google.com/icons) or [here](https://materialdesignicons.com)?

ahmed-fathy-aly:
*updated*
Thanks @xiprox 
I applied your tips and updated the screenshot in the description

Now all the icons come from https://design.google.com/icons/ 
I agree with you, Anyone is welcome to replace them with a better custom mades one in the future. Same for the header background, now I am using the generic one used in the material design guidelines.

@ishan1604  are we all set up for the merge ?

xiprox:
Great job! Looks much more polished now. Are the Dashboard and Client icons still small though?

Another thing that I just thought about: You might want to set `android:lines="1"` and `android:ellipsize="end"` to `tv_user_name` in order to avoid any possible glitches caused by long names.

ahmed-fathy-aly:
*Updated*
- increased the dashboard icon's size.
- increased the clients icon's size a little bit. The icon itself is a bit smaller than the groups icon, but I think it looks better that way with respect to the relative size of the silhouettes in both icons(if both icons were the same size the clients' silhouettes would look much bigger than the groups').
- Restricted a really long name to be in one line ending with "..." instead of spreading on multiple lines. 

Thanks again @xiprox

xiprox:
Well, the standard is 24dp but it's ok, I guess. This looks all good to me! No issues, and great job once again. 👍

ahmed-fathy-aly:
@ishan1604  Can we merge it now ?

PR #246, opened 2016-04-03
Survey New Feature Added with Some Enhancement.

therajanmaurya:
**New Features added In Survey and some feature enhanced**

1. A subtitle on the toolbar that tells the count of current question / total questions.
2. Provided one submit button at the end of the survey.
3. The survey responses will be persisted all the time (in memory) and user can change them anytime they want till they have not pressed the submit button.
4. User can move back and forth in a survey onto any question. (Non Swipeable View Pager is replaced with default one).
5. The app is persisting the last state if for any reason app goes to the background while filling up the survey. 

and **Squashed**.


ishan1604:
Great Job @therajanmaurya !!

Issue #190, opened 2016-03-09
NullPointerException when coming back from Add savings account page.


Sreedev:
- Click a client from client list.
- From toolbar select Add Savings Account.
- Without entering any data just press back button.
Sometimes app crashed, sometimes not.

03-09 12:06:27.364  22635-22635/com.mifos.mifosxdroid E/AndroidRuntime﹕ FATAL EXCEPTION: main
    Process: com.mifos.mifosxdroid, PID: 22635
    java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Object android.content.Context.getSystemService(java.lang.String)' on a null object reference
            at android.view.LayoutInflater.from(LayoutInflater.java:229)
            at android.widget.ArrayAdapter.<init>(ArrayAdapter.java:178)
            at android.widget.ArrayAdapter.<init>(ArrayAdapter.java:163)
            at com.mifos.mifosxdroid.online.SavingsAccountFragment$6.success(SavingsAccountFragment.java:486)
            at com.mifos.mifosxdroid.online.SavingsAccountFragment$6.success(SavingsAccountFragment.java:444)
            at retrofit.CallbackRunnable$1.run(CallbackRunnable.java:45)
            at android.os.Handler.handleCallback(Handler.java:739)
            at android.os.Handler.dispatchMessage(Handler.java:95)
            at android.os.Looper.loop(Looper.java:148)
            at android.app.ActivityThread.main(ActivityThread.java:5417)
            at java.lang.reflect.Method.invoke(Native Method)
            at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
            at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)

Sreedev:
I am working on this.

RowlandOti:
@Sreedev , have you solved this?

Sreedev:
@RowlandOti  No, You can take this forward.

Sreedev:
@RowlandOti  Sorry for late reply. I am out of station, you can work on this.

xiprox:
I think I see what the issue here is. The Fragment is probably no longer attached to an Activity by the time the `success` method is executed. This causes `null` to be passed to the `ArrayAdapter` which in turn causes the `java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Object android.content.Context.getSystemService(java.lang.String)' on a null object reference` exception to be thrown. I'll look into it if it's ok.

[...]

PR #255, opened 2016-04-07
MIFOSAC-139: Improve network operation progress indication

xiprox:
At the moment, progress indication for network operations is done by displaying Progress Dialogs. This is bad for al least two reasons:
**1)** It's bad UX as it heavily restricts users. Users have no means of navigating back from a Fragment/Activity until an operation completes. And since these are network operations, their duration may greatly vary from connection to connection, and sometimes cause intolerable wait times. This could be extremely annoying if a user accidentally navigates into such a situation.
**2)** In places where multiple operations are executed in a queue, we see multiple "blinks" and dialogs popping up one after the other. This is pretty ugly.

As a solution, we could display `ProgressBar`s inside layouts themselves with the help of the `ViewFlipper` class. An example of how this looks can be found here: https://streamable.com/nnes

#### Roadmap:
✔️  **1.** Add [`ProgressableFragment`](https://github.com/xiprox/android-client/blob/progress_improvements/mifosng-android/src/main/java/com/mifos/mifosxdroid/core/ProgressableFragment.java) and  [`ProgressableDialogFragment`](https://github.com/xiprox/android-client/blob/progress_improvements/mifosng-android/src/main/java/com/mifos/mifosxdroid/core/ProgressableDialogFragment.java) base classes which extend `MifosBaseFragment` or `DialogFragment` and provides a method (`showProgress(boolean)`) for easy `ViewFlipper` flipping.

✔️  **2.** Make the following Classes/Fragments extend `ProgressableFragment` or `ProgressableDialogFragment` (if Fragments or DialogFragments), replace `ProgressDialog` calls with `ViewFlipper` calls, and make the necessary changes to their layouts (addition of `ViewFlipper` and `ProgressBar`):
- [x] CenterListFragment
- [x] ClientDetailsFragment
- [x] LoanAccountSummaryFragment
- [x] SavingsAccountSummaryFragment
- [x] GroupDetailsFragment
- [x] CreateNewClientFragment
- [x] CreateNewGroupFragment
- [x] ClientChooseFragment
- [x] CenterListFragment
- [x] SurveyListFragment
- [x] GroupListFragment
- [x] ClientIdentifiersFragment
- [x] DataTableDataFragment
- [x] DocumentListFragment
- [x] GenerateCollectionSheetFragment
- [x] LoanRepaymentFragment
- [x] LoanRepaymentScheduleFragment
- [x] SavingsAccountTransactionFragment
- [x] ~~SavingsAccountTransactionFragment#processTransaction~~  - We are making a transaction here. We could implement a Snackbar-based solution here. We'd have to implement operation cancelation too in that case though. I'm simply leaving this untouched.
- [x] GroupLoanAccountFragment
- [x] ~~AsyncFileDownloader~~ - A better solution could be a Snackbar or a Notification. That'd require too much work and the implementation of cancellability though. Leaving untouched.
- [x] ChargeDialogFragment - There are two such calls here. One is for loading Spinner entries and the second is for the save operation. I'm replacing the load ProgressDialog with ViewFlipper. The dialog for the save operation stays there though.
- [x] ~~DataTableRowDialogFragment~~ - Not able to test. Looks like an operation initiated by a button click. Better off with a Progress Dialog.
- [x] ~~DocumentDialogFragment~~ - User initiated Upload operation. Would require work on cancellability. Better off with a Progress Dialog.
- [x] LoanAccountFragment
- [x] LoanChargeDialogFragment
- [x] SavingsAccountFragment

**Note about ClientListFragment and GroupListFragment:** These two Fragments implement the SwipeRefresh pattern. Currently, the progress indicators aren't being shown on first load. They will be omitted in this PR and fixes for the SwipeRefreshLayouts' indicators will be submitted as a separate PR.

✔️  **3.** Fix crashes caused by not keeping a reference to a `Context` and directly calling `getActivity()` in async operation results. Fragments get detached from their parent `Activity` and the `getActivity()` call returns `null`. A call on this `null` is made at a further point which in turn causes an exception to be thrown. The reason this never came up until now is because Progress Dialogs prevented users from leaving a Fragment before the result of an async operation was delivered. Fix is easy; just keep a reference to the parent `Activity`/`Context` from when it isn't `null`. An example of this can be seen in #190.

### DIFF
--- a/mifosng-android/src/main/java/com/mifos/mifosxdroid/core/ProgressableDialogFragment.java
+++ b/mifosng-android/src/main/java/com/mifos/mifosxdroid/core/ProgressableDialogFragment.java
@@ -1,0 +1,48 @@
+package com.mifos.mifosxdroid.core;
+
+import android.os.Bundle;
+import android.support.annotation.Nullable;
+import android.support.v4.app.DialogFragment;
+import android.util.Log;
+import android.view.View;
+import android.widget.ProgressBar;
+import android.widget.ViewFlipper;
+
+import com.mifos.mifosxdroid.R;
+
+/**
+ * A {@link android.support.v4.app.DialogFragment} that provides progress viewing functionality with a ViewFlipper.
+ */
+public class ProgressableDialogFragment extends DialogFragment {
+    public static final int VIEW_FLIPPER_PROGRESS = 0;
+    public static final int VIEW_FLIPPER_CONTENT = 1;
+
+    private ViewFlipper viewFlipper;
+
+    @Override
+    public void onViewCreated(View view, @Nullable Bundle savedInstanceState) {
+        super.onViewCreated(view, savedInstanceState);
+
+        viewFlipper = (ViewFlipper) view.findViewById(R.id.view_flipper);
+
+        if (viewFlipper == null) {
+            throw new NullPointerException("Are you sure your Fragment has a ViewFlipper with id \"view_flipper\"?");
+        }
+    }
+
+    /**
+     * Switches the {@link ViewFlipper} either to the {@link ProgressBar} or to the content.
+     *
+     * @param show true to show {@link ProgressBar}, false for content.
+     */
+    public void showProgress(boolean show) {
+        try {
+            int childToFlipTo = show ? VIEW_FLIPPER_PROGRESS : VIEW_FLIPPER_CONTENT;
+            if (viewFlipper.getDisplayedChild() != childToFlipTo) {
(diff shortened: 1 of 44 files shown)


xiprox:
Alright! This PR is now complete and ready to be merged.

Do you need me to squash these three commits as well or are they fine the way they are? Feels like it might be better to keep them separated like this? If not, to squash 856b6eb and d2938f7 but to leave 3abacaf separate?

PR #256, opened 2016-04-08
MIFOSX-142 Unnecessary search button on activities related to center list resolved

mayank-kgp:
There was a unnecessary search button on center list page which closes the activity on pressing. 
Moreover successive fragment has also these button - 
1. Fragment that contain the list of groups in any particular center
2. Client list page that contain the list of client in any particular group
Now there is no search button.

<a href="url"></a>

<a href="url"></a>

<a href="url"></a>


mayank-kgp:
@ishan1604  Please take a review of it.

ishan1604:
Thanks for the PR but this was already fixed in PR#250

PR #265, opened 2016-04-13
Added client list associated with a group

mayank-kgp:
I have added functionality of viewing clients associated with a group.
Moreover if no client is associated, fragment automatically closed after showing a toast.

<a href="url"></a>
<a href="url"></a>
<a href="url"></a>


mayank-kgp:
@ishan1604   Please have a look on this.

PR #268, opened 2016-04-14
add the ability to create center savings account

ahmed-fathy-aly:
MIFOSAC-129
- creates a savings account for a center
- added an activity test about accessing the SavingsAccountFragment from the
  CentersActivity

Adding a savings account is the same for clients, groups or centers
The only difference is in the SavingsPayload object being sent
So I used the factory design pattern to let us use the SavingsAccountFragment for the three types(client, group, center) as follows : 
- SavingsAccountFragment receives the clientType in the arguments
- when submitting, SavingsAccountFragment would ask SavingsPayloadFactory to create the SavingsPayload corresponding to the clientType it received in the arguments

![Screenshot]
(http://i.imgur.com/g7Q3oiz.png)


ahmed-fathy-aly:
@ishan1604  
Any problems with this PR ?

PR #270, opened 2016-04-16
ClientListFragment and ClientSearchFragment Ui test with espresso latest documentation

therajanmaurya:
Fragmenttransaction in Navigation Drawer. 
ClientListFragment ListView -> RecyclerView.
Standard Navigation Drawer action added in test. 
ClientSearchFragment Ui test.
ClientListFragment Ui test.
Espresso full setup done.

and **Squashed Done**


ahmed-fathy-aly:
Is it the right thing to have a test that compares that data loaded with static ones like you did in ClientListFragmentTest ?
If the real data was edited or another instance URL was used the test would fail.

ahmed-fathy-aly:
And the same for ClientSearchFragmentTest
It searches for a client called "JAVA BOY" and assert the id of the search result to start with "#10"
If the client's name was edited, the test case would fail.

therajanmaurya:
ClientListFragmentTest -> Ui test written in that manner, when project architecture implementation done so we need only set the present fake data to the fragment and do not need to change in the test. it will pass always.

ClientSearchFragment -> it is now fix. in my new commit and it is squashed now.

Thanks for suggestion.

ahmed-fathy-aly:
ClientSearchFragmentTest is better now, Good job.

For ClientListFragmentTest, Whenever a developer wants to run that test he should update CLIENTLIST_JSON with the present data from the backend he's using, right ?

therajanmaurya:
No,They do not need to update the CLIENTLIST_JSON because when  project architecture implementation done so we need only set the present fake data(CLIENTLIST_JSON) to the ClientListFragment view layer and do not need to change in the test. it will pass always.

PR #271, opened 2016-04-17
Fix SwipeRefreshLayout indicators not being shown

xiprox:
...in ClientListFragment and GroupsListFragment on initial fetch.

As mentioned in #255:
> **Note about ClientListFragment and GroupListFragment:** These two Fragments implement the SwipeRefresh pattern. Currently, the progress indicators aren't being shown on first load. They will be omitted in this PR and fixes for the SwipeRefreshLayouts' indicators will be submitted as a separate PR.


xiprox:
Rebased.

PR #276, opened 2016-04-26
MIFOSAC-127: Add ability to add charges to a savings account

chegz254:
Copied working code from old branch to new branch. The previous branch contained code for other tasks, which interfered with the main code.


mifosng-android/src/main/java/com/mifos/mifosxdroid/dialogfragments/SavingsChargeDialogFragment.java
```diff
@@ -0,0 +1,245 @@
...
+                final ArrayAdapter<String> chargesAdapter = new ArrayAdapter<String>(getActivity(),
+                        android.R.layout.simple_spinner_item, chargesNames);
+                chargesAdapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
+                sp_charge_name.setAdapter(chargesAdapter);
+                sp_charge_name.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
+                    @Override
+                    public void onItemSelected(AdapterView<?> adapterView, View view, int i, long l) {
+                        Id = chargeNameIdHashMap.get(chargesNames.get(i));
+                        Log.d("chargesoptionss" + chargesNames.get(i), String.valueOf(Id));
+                        if (Id != -1) {
```
ishan1604: Could you please explain the reason for this empty block?

chegz254: That block of code deals with loading the savings charges options into the spinner for selecting charge options using the spinner.

I tested using the domain of demo.openmf.org with the account of  John Kwame Doe (Client Id 201). I was able to load the four charge options using JSON and selected one charge from the spinner.

ishan1604: Ohh I see that. But I don't understand why is the if block is empty ? and you are doing something in the else block. Can't u inverse the logic and say if (Id == -1) { do your stuff }}

chegz254: In the else block, a Toast notification will appear when no charge is available or has been selected. The if block will check if the charge selected is greater than 0. Also that Id represents the chargeId for a selected charge. If it is alright, I shall make a new commit with that change.

PR #299, opened 2016-06-23
Feature Offline 

therajanmaurya:
- [X]  setup DBFlow
- [X]  Designing Architecure for offline (DatabaseHelper Layer)
- [X]  Making DBFlow Model for ClientList


therajanmaurya:
Hi, @fomenkoo can you help me on some point, How did you think to make the offline feature and some other staff login etc.
Thanks in advance

PR #310, opened 2016-07-23
Offline Clients and Groups Sync with Server Error Handling

therajanmaurya:
Full Functionality during Sync the Client and Groups
- [x]  Sync Clients
- [x] Sync Groups
- [x] Error Handling If Clients or Groups  are unable to create. 
Example With this External Id already a client exists etc.


therajanmaurya:
Looks I updated the Gradle tools Version 1.3.0 to 2.10 Latest.
In which many changes have made in Gradle version 2.10, So need to update fabric crashlytics version.
Reference : http://stackoverflow.com/questions/32959477/android-studio-gradle-build-fails-in-release-could-not-determine-the-dependen

PR #319, opened 2016-08-15
fix #316

### LINKED ISSUE #316: Delete an identifier
After Deleting an identifier it still shows the same identifier, instead it should be removed.
Since its there clicking on remove again gives a 404

Steps to Reproduce:
- Login
- Clients
- Select a client (with an identifier)
- Select Identifiers from the dropdown menu
- Click Remove

chashmeetsingh:


ishan1604:
@chashmeetsingh Thanks for sending the PR. 

Here are a few things needed before I can review the PR:
- The commits must be squashed before sending the PR.
- A proper description about the PR (What it fixes/improves, How it does that?)
- Reference of any Github Issue IDs that are linked to it.

PR #320, opened 2016-08-15
Offline sync on muti selection

therajanmaurya:
- [x]  Sync sign added that show Client Synced or not.
- [x]  RecyclerView MultiSelection Added.
- [x]  Select Client and Sync.
- [x]  Final User is able to select and sync clients at the single place.


ishan1604:
Build is failing.

PR #324, opened 2016-08-15
fix identifier delete issue and add swipe to refresh

chashmeetsingh:
- Removes identifier from view after deleted from the API request
- Fixes issue #316
- Added swipe to refresh

Steps:
- Listener for the Remove button
- Its onClick makes an API call to delete the identifier
- On successful response identifier List updated and `notifyDataSetChanged()` called


chashmeetsingh:
thanks! @ishan1604

ishan1604:
Thanks to you @chashmeetsingh for the good work. Keep the PRs flowing 👍

PR #329, opened 2016-08-20
Offline Sync SavingsAccountTransaction

therajanmaurya:

Adapter added and basic saving Account transaction Fragment added with presenter and view

Offline Sync SavingsAccountTransaction Feature added.
Offline Sync  SavingsAccountTransaction  and SavingsAccountsTemplate On Muti-Selection
Documentation added

### DIFF
--- a/mifosng-android/src/main/java/com/mifos/mifosxdroid/offline/syncsavingsaccounttransaction/SyncSavingsAccountTransactionFragment.java
+++ b/mifosng-android/src/main/java/com/mifos/mifosxdroid/offline/syncsavingsaccounttransaction/SyncSavingsAccountTransactionFragment.java
@@ -1,0 +1,234 @@
+package com.mifos.mifosxdroid.offline.syncsavingsaccounttransaction;
+
+import android.content.DialogInterface;
+import android.os.Bundle;
+import android.support.v4.widget.SwipeRefreshLayout;
+import android.support.v7.widget.LinearLayoutManager;
+import android.support.v7.widget.RecyclerView;
+import android.view.LayoutInflater;
+import android.view.Menu;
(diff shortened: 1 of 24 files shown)


mifosng-android/src/main/java/com/mifos/mifosxdroid/offline/syncsavingsaccounttransaction/SyncSavingsAccountTransactionFragment.java
```diff
@@ -0,0 +1,233 @@
...
+    /**
+     * Loading All SavingsAccount Transactions from Database On SwipeRefresh
+     */
+    @Override
+    public void onRefresh() {
+        //Loading LoanRepayment Transactions and PaymentTypeOptions From Database
+        mSyncSavingsAccountTransactionPresenter.loadDatabaseSavingsAccountTransactions();
+        mSyncSavingsAccountTransactionPresenter.loadPaymentTypeOption();
+
+        if (swipeRefreshLayout.isRefreshing())
```
ishan1604: Problem here!

write the statement to be executed inside the if block on the same line if you are not using braces.
This way you prevent people and yourself from adding any code below it that might end up being confused with the if block

ishan1604:
The Code Seems to be fine and I am going to merge it. 

One small thing that needs to be taken care of is that you are writing single line if blocks with a new line.

How we can solve it?
Let's use one of the code quality plugins and add a check there that fails the build when we add if - else block with body on the next line without braces.

What do you think ?

therajanmaurya:
Yes, it's fine. I think, It will be good to add in code quality plugins for checking **if - else block with body on the next line without braces.**

I will add this in next PR.

Thanks

PR #341, opened 2016-09-19
Feature Offline Sync groups

therajanmaurya:

GroupList Enhance.
GroupDetails API Request Enhance.
Offline Sync Group Details and their accounts.
Database Encryption added.


mifosng-android/src/main/java/com/mifos/api/datamanager/DataManagerGroups.java
```diff
@@ -52,14 +53,7 @@ public DataManagerGroups(BaseApiManager baseApiManager,
...
             case 0:
-                return mBaseApiManager.getGroupApi().getGroups(paged, offset, limit)
-                        .concatMap(new Func1<Page<Group>, Observable<? extends Page<Group>>>() {
-                            @Override
-                            public Observable<? extends Page<Group>> call(Page<Group> groupPage) {
-                                mDatabaseHelperGroups.saveGroups(groupPage);
-                                return Observable.just(groupPage);
-                            }
-                        });
+                return mBaseApiManager.getGroupApi().getGroups(paged, offset, limit);
```
ishan1604: Good work pulling this code into a separate method.

mifosng-android/src/main/java/com/mifos/api/local/MifosSQLCipherHelper.java
```diff
@@ -0,0 +1,18 @@
...
+public class MifosSQLCipherHelper extends SQLCipherOpenHelper {
+
+    public MifosSQLCipherHelper(DatabaseDefinition databaseDefinition,
+                                DatabaseHelperListener listener) {
+        super(databaseDefinition, listener);
+    }
+
+    @Override
+    protected String getCipherSecret() {
+        return "dbflow-rules";
```
ishan1604: I don't know how this CipherSecret works in DBFlow but don't you think we should generate it differently for each App Install.

Let's say we have a function that takes certain inputs and returns an output. For a given input the output should always be the same. 

Here is how it would look

SOME VALUE XYZ which is unique per device (Android user) ----> Function -----> Output is some hash

use this Hash as a Cipher Secret. 

So that it becomes even more difficult to decode the encrypted DB. 

Makes sense ?

mifosng-android/src/main/java/com/mifos/api/local/databasehelper/DatabaseHelperGroups.java
```diff
@@ -34,50 +38,134 @@ public DatabaseHelperGroups() {
...
+                        .from(Group.class)
+                        .where(Group_Table.id.eq(groupId))
+                        .querySingle();
+
+                if (group != null) {
+                    group.setActivationDate(Arrays.asList(group.getGroupDate().getDay(),
+                            group.getGroupDate().getMonth(), group.getGroupDate().getYear()));
+                }
+
+                return null;
```
ishan1604: Why are we returning null here ?

mifosng-android/src/main/java/com/mifos/api/local/MifosSQLCipherHelper.java
therajanmaurya: Yeah I Agree. Can I do this in separate PR.

mifosng-android/src/main/java/com/mifos/api/local/databasehelper/DatabaseHelperGroups.java
therajanmaurya: Sorry for that When I have written this method, demo Server was down, So I have not test this method, Now fixed.

mifosng-android/src/main/java/com/mifos/api/local/MifosSQLCipherHelper.java
ishan1604: Sure

ishan1604: If we are to add it in a separate PR, do all the Cipher work together because if it gets squashed in this one it will be difficult to track in future. so remove the cipher related code from this pr.

therajanmaurya: Ok, I agree. I will do that.

therajanmaurya: It's done @ishan1604. I have removed the SQLchiper code and now everything good. you can merge.

mifosng-android/src/main/java/com/mifos/api/local/databasehelper/DatabaseHelperGroups.java
therajanmaurya: It is fixed in below commit.

PR #379, opened 2016-10-10
Image downloading through glide instead of async task

top-gun007:

### DIFF
--- a/mifosng-android/src/main/java/com/mifos/mifosxdroid/online/clientdetails/ClientDetailsFragment.java
+++ b/mifosng-android/src/main/java/com/mifos/mifosxdroid/online/clientdetails/ClientDetailsFragment.java
@@ -8,11 +8,9 @@ package com.mifos.mifosxdroid.online.clientdetails;
 import android.app.Activity;
 import android.content.Intent;
 import android.content.IntentSender;
-import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.location.Location;
 import android.net.Uri;
-import android.os.AsyncTask;
 import android.os.Bundle;
 import android.provider.MediaStore;
 import android.support.v4.app.FragmentTransaction;
@@ -37,6 +35,9 @@ import android.widget.TableRow;
 import android.widget.TextView;
 import android.widget.Toast;
 
+import com.bumptech.glide.Glide;
+import com.bumptech.glide.load.model.GlideUrl;
+import com.bumptech.glide.load.model.LazyHeaders;
 import com.google.android.gms.common.ConnectionResult;
 import com.google.android.gms.common.api.GoogleApiClient;
 import com.google.android.gms.location.LocationListener;
@@ -70,11 +71,6 @@ import com.mifos.utils.PrefManager;
 import com.mifos.utils.Utils;
 
 import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.net.HttpURLConnection;
-import java.net.MalformedURLException;
-import java.net.URL;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
@@ -113,6 +109,7 @@ public class ClientDetailsFragment extends ProgressableFragment implements Googl
 
     public int clientId;
     public List<DataTable> clientDataTables = new ArrayList<>();
+    public String urlStringBuilder;
     List<Charges> chargesList = new ArrayList<Charges>();
 
     @BindView(R.id.tv_fullName)
@@ -169,7 +166,6 @@ public class ClientDetailsFragment extends ProgressableFragment implements Googl
     private AtomicBoolean locationAvailable = new AtomicBoolean(false);
 
     private AccountAccordion accountAccordion;
-    private ImageLoadingAsyncTask imageLoadingAsyncTask;
 
 
     /**
@@ -184,14 +180,6 @@ public class ClientDetailsFragment extends ProgressableFragment implements Googl
         args.putInt(Constants.CLIENT_ID, clientId);
         fragment.setArguments(args);
         return fragment;
-    }
-
-    @Override
-    public void onDetach() {
-        if (imageLoadingAsyncTask != null && !imageLoadingAsyncTask.getStatus().equals(AsyncTask
-                .Status.FINISHED))
-            imageLoadingAsyncTask.cancel(true);
-        super.onDetach();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


top-gun007:
@therajanmaurya please have a look

mifosng-android/src/main/java/com/mifos/mifosxdroid/online/clientdetails/ClientDetailsFragment.java
```diff
@@ -599,6 +600,26 @@ public void showUploadImageProgressbar(boolean b) {
         }
     }
 
+
+    public void loadImageGlide(int clientId) {
```
top-gun007: - this method is called in all instances instead of async task.
- creates a glide url by adding url with the header as provided.
- provoked in two methods i.e.- showUploadImageFailed(String s) and showClientInformation(Client client)

top-gun007:
Chages -created a method loadImageGlide(int) which is called instead of ImageLoadingAsyncTask
- creates a glide url by adding url with the header as provided.
- provoked in two methods i.e.- showUploadImageFailed(String s) and showClientInformation(Client client)

ishan1604:
Please squash the PR before asking for review. 

Squashed commits allow us to review all the changes at one in each file, instead of visiting each commit for each file. 

Here is how you can squash

```
git rebase -i HEAD~n
```

-i stands for interactive rebase
n is the number previous commits you want to rebase, this should be the number of commits you want to squash.

Now when you see the git rebase editor open up in your editor just write **squash** in front of each commit you want to squash except the first commit i.e. the latest one.

mifosng-android/src/main/java/com/mifos/mifosxdroid/online/clientdetails/ClientDetailsFragment.java
```diff
@@ -599,6 +600,26 @@ public void showUploadImageProgressbar(boolean b) {
...
+        pb_imageProgressBar.setVisibility(VISIBLE);
+        String url = PrefManager.getInstanceUrl()
```
ishan1604: For every client detail you are creating a new String URL instance, that too with concatenation which is not a very good idea.

Consider using StringBuilder instead.

Also try to abstract this out into some util and cache the URL with clientID placeholder. So that you only have to insert that later on.

So your cached your would look something like this "blah blah %s blah blah blah"

And now every time you request this url from your Util you have to insert clientId to complete that.

mifosng-android/src/main/java/com/mifos/mifosxdroid/online/clientdetails/ClientDetailsFragment.java
```diff
@@ -599,6 +600,26 @@ public void showUploadImageProgressbar(boolean b) {
...
+                + "clients/"
+                + clientId
+                + "/images?maxHeight=120&maxWidth=120";
+
+        GlideUrl glideUrl = new GlideUrl(url, new LazyHeaders.Builder()
```
ishan1604: Similarly, you can also take this glideURL thing out of it and put it in the same class that gives you the String URL. So basically, you would never want to expose the URL at first and what should be exposed via a public method is the glideURL to the view that wants to show the Image.

Please let me know if you didn't understand any part of the review, I'd be happy to help.

mifosng-android/src/main/java/com/mifos/mifosxdroid/online/clientdetails/ClientDetailsFragment.java
```diff
@@ -600,15 +604,21 @@ public void showUploadImageProgressbar(boolean b) {
...
 
+    public void urlBuilder() {
+        url = new StringBuilder();
+        url.append(PrefManager.getInstanceUrl()
+                + "clients/"
+                + "clientId"
+                + "/images?maxHeight=120&maxWidth=120");
+        String tempUrlCopy = url.toString();
+        tokenPositionStart = tempUrlCopy.indexOf("clientId");
+        tokenPositionEnd = tokenPositionStart + 8;
```
ishan1604: Good attempt abstracting out this method. 

You can do this simply like this

```
sb = StringBuilder(PrefManager.getInstanceUrl());
sb.append(String.format("clients/%d/images?maxHeight=120&maxWidth=120",clientId);
String url = sb.toString();
```

You don't need to do tokenPositionStart,End calculations and then a replace below.

top-gun007: ok @ishan1604 I will but can you help in figuring out why the checks are not passing on  my pr as I have only changed the way "url" is constructed as compared to my previous PR.
Thank you

top-gun007: ok @ishan1604 I will but can you help in figuring out why the checks are not passing on  my pr as I have only changed the way "url" is constructed as compared to my previous PR.
Thank you

top-gun007: thanks @mayank-kgp previously using ./gradlew connectedCheck

mifosng-android/src/main/java/com/mifos/mifosxdroid/online/clientdetails/ClientDetailsFragment.java
```diff
@@ -113,6 +109,7 @@
 
     public int clientId;
     public List<DataTable> clientDataTables = new ArrayList<>();
+    public String sb;
```
therajanmaurya: ambiguous name, this can be urlStringBuilder, @ishan1604 what you think about that.
I think rest of thing looks good, Please give final look to this PR.

mifosng-android/src/main/java/com/mifos/mifosxdroid/online/clientdetails/ClientDetailsFragment.java
```diff
@@ -188,9 +184,6 @@ public static ClientDetailsFragment newInstance(int clientId) {
 
     @Override
     public void onDetach() {
```
therajanmaurya: If this method is no longer in use, please remove.

mifosng-android/src/main/java/com/mifos/mifosxdroid/online/clientdetails/ClientDetailsFragment.java
ishan1604: Agreed with @therajanmaurya

therajanmaurya:
@ishan1604 This PR looks good. Do the honors.

Issue #389, opened 2016-11-05
No Response for Invalid username/password in login activity


chiragw15:
When a user enters wrong username or password, there is no response.

**Expected behavior:** 

There should be a message if user enters invalid username/password. 

These are some implementations, I tried.

1.) SnackBar
![screenshot_20161105-220253](https://cloud.githubusercontent.com/assets/16978820/20032088/e540ba68-a3a8-11e6-9b3c-4a5fbca24dc0.png)

2.)Toast
![screenshot_20161105-214433](https://cloud.githubusercontent.com/assets/16978820/20032089/f0df2134-a3a8-11e6-9446-8e0751cb3d4d.png)

3.) DialogBox
![screenshot_20161105-220901](https://cloud.githubusercontent.com/assets/16978820/20032093/ffb30194-a3a8-11e6-8136-03018ccc8e72.png)

@therajanmaurya Please let me know which implementation looks good.

therajanmaurya:
The first approach looks good 👍

chiragw15:
Okay sure. I'll be sending a PR for this soon.

therajanmaurya:
fixed in #522

tarun0:
@iadeelzafar Create a new issue for that. Make sure to add a gif to reproduce the issue.

PR #393, opened 2016-11-11
Correcting Login Activity

chiragw15:

Things implemented in the PR:
1.) Added code to display error if username/password is incorrect as mentioned in issue #389 . 
2.) Corrected the code to display error if user is not connected to internet.
3.) Previously, app crashed on entering invalid instance url. Fixed that.
4.) Pressing exit button on SSL handshake dialog box would exit the app. Corrected it to exit dialog box instead.
5.) Icon on dialog box was not getting displayed. Corrected it.
6.) The colour of instance url always remained green whether the user has entered correct url or not.    Corrected it. Now if the user has entered wrong url, it displays instance url in red colour.
7.) Moved all hard-coded strings in Login Activity to strings.xml and also added their translations.
8.) Removed all extra imports.

### DIFF
--- a/mifosng-android/src/main/java/com/mifos/mifosxdroid/login/LoginActivity.java
+++ b/mifosng-android/src/main/java/com/mifos/mifosxdroid/login/LoginActivity.java
@@ -9,7 +9,6 @@ import android.app.AlertDialog;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Bundle;
-import android.support.v4.content.ContextCompat;
 import android.text.Editable;
 import android.text.InputType;
 import android.text.TextWatcher;
@@ -21,7 +20,6 @@ import android.widget.LinearLayout;
 import android.widget.TextView;
 
 import com.mifos.api.BaseApiManager;
-import com.mifos.api.BaseUrl;
 import com.mifos.mifosxdroid.R;
 import com.mifos.mifosxdroid.core.MifosBaseActivity;
 import com.mifos.mifosxdroid.core.util.Toaster;
@@ -77,7 +75,8 @@ public class LoginActivity extends MifosBaseActivity implements LoginMvpView {
     private String username;
     private String instanceURL;
     private String password;
-    private boolean isValidUrl;
+    private String domain;
+    private boolean isValidUrl = false;
 
     private TextWatcher urlWatcher = new TextWatcher() {
         @Override
@@ -94,15 +93,20 @@ public class LoginActivity extends MifosBaseActivity implements LoginMvpView {
         public void afterTextChanged(Editable editable) {
             Integer port = et_port.getEditableText().toString().isEmpty() ? null : Integer
                     .valueOf(et_port.getEditableText().toString());
+
             instanceURL = ValidationUtil.getInstanceUrl(et_domain.getText().toString(), port);
             isValidUrl = ValidationUtil.isValidUrl(instanceURL);
-            tv_full_url.setText(instanceURL);
-            tv_full_url.setTextColor(isValidUrl ?
-                    ContextCompat.getColor(getApplicationContext(),
-                            R.color.green_light) :
-                    ContextCompat.getColor(getApplicationContext(),
-                            R.color.red_light));
 
+            tv_full_url.setText(instanceURL);
+
+            domain = et_domain.getEditableText().toString();
+
+            if (domain.length() == 0 || domain.contains(" ")) {
+                isValidUrl = false;
+            }
+
+            tv_full_url.setTextColor(getResources()
+                    .getColor(isValidUrl ? R.color.green_light : R.color.red_light));
         }
     };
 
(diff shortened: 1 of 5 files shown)


mifosng-android/src/main/java/com/mifos/mifosxdroid/login/LoginActivity.java
```diff
@@ -136,18 +140,27 @@ public void onClick(View view) {
     }
 
     public boolean validateUserInputs() {
+
+        domain = et_domain.getEditableText().toString();
+        if (domain.length() == 0 || domain.contains(" ")) {
+            Toaster.show(findViewById(android.R.id.content), getString(R.string.error_invalid_url));
```
therajanmaurya: It's looks odd, make a separate method for showing Toast. Like this
```
public void showToast(String message) {
      Toaster.show(context, message); // I think "this" will work in place of context
}
```

mifosng-android/src/main/java/com/mifos/mifosxdroid/login/LoginActivity.java
```diff
@@ -171,32 +184,36 @@ public void onLoginSuccessful(User user) {
...
-            }*/
-        } catch (NullPointerException e) {
-            if (Network.getConnectivityStatusString(this).equals("Not connected to " +
-                    "Internet")) {
-                Toaster.show(findViewById(android.R.id.content), "Not connected to Network");
-            } else {
-                Toaster.show(findViewById(android.R.id.content), getString(R.string.error_unknown));
-            }
+
+        if (throwable instanceof SSLHandshakeException) {
```
therajanmaurya: We are using Retrofit 2, you can cast ```Throwable``` with ```HttpException``` and access failed errors.
For example look here https://github.com/openMF/android-client/blob/master/mifosng-android/src/main/java/com/mifos/mifosxdroid/dialogfragments/identifierdialog/IdentifierDialogPresenter.java#L93

mifosng-android/src/main/java/com/mifos/mifosxdroid/login/LoginActivity.java
```diff
@@ -210,27 +227,34 @@ public void showProgressbar(boolean show) {
      */
     private void promptUserToByPassTheSSLHandshake() {
         new AlertDialog.Builder(this)
-                .setTitle("SSL Certificate Problem")
-                .setMessage("There is a problem with your SSLCertificate, would you like to " +
-                        "continue? This connection would be unsafe.")
-                .setIcon(android.R.drawable.stat_sys_warning)
-                .setPositiveButton("Continue", new DialogInterface.OnClickListener() {
+                .setTitle(getString(R.string.ssl_certificate_dialog))
```
therajanmaurya: Looks good, Please remove default ```AlertDialog``` with this one https://github.com/openMF/android-client/blob/master/mifosng-android/src/main/java/com/mifos/mifosxdroid/core/MaterialDialog.java

mifosng-android/src/main/java/com/mifos/mifosxdroid/login/LoginActivity.java
```diff
@@ -240,16 +264,7 @@ public void onLoginClick(Button button) {
...
-            BaseApiManager.createService();
-        }
-
-        login(false);
-    }
-
-    private void login(boolean shouldByPassSSLSecurity) {
-        if (!validateUserInputs())
-            return;
+        BaseApiManager.createService();
```
therajanmaurya: ```if (!PrefManager.getInstanceDomain().equals(BaseUrl.API_ENDPOINT))``` this line telling us that user have changed the base URL or not, if not then we not need to do any task, do not need to write again base URL in Preference.

mifosng-android/src/main/res/values/strings.xml
```diff
@@ -453,6 +463,8 @@
     <string name="dialog_action_i_am_sure">I\'M Sure</string>
     <string name="dialog_action_re_try">Re-Try</string>
     <string name="dialog_action_app_settings">App Settings</string>
+    <string name="Continue">Continue</string>
```
therajanmaurya: ```<string name="Continue">Continue</string>``` this should be <string name="dialog_action_continue">Continue</string>

therajanmaurya:
@chiragw15 Please make the above changes

mifosng-android/src/main/java/com/mifos/mifosxdroid/login/LoginActivity.java
therajanmaurya: I am talking about the recreating the service, not about the 
```
if (!validateUserInputs())
           return;
```

chiragw15: I am also talking about recreating service. Suppose, I changed Base URL to www.chirag.com. Therefore, InstanceDomain in PrefManager will be changed to "www.chirag.com". Now, since calling on this website for api will generate an exception. Hence user will not be able to login.
Now I again changed the Base URL to the default one i,e demo.openmf.org. Now, InstanceDomain in PrefManager will be changed to "demo.openmf.org". So, now since InstanceDomain in PrefManager == BaseUrl.API_ENDPOINT , the BaseApiManager.createService() won't be called and hence user won't be able to login again.

therajanmaurya: Yeah, I got it, Sorry I have missed one case, Thanks go ahead for the rest of things.

mifosng-android/src/main/java/com/mifos/mifosxdroid/login/LoginActivity.java
chiragw15: @therajanmaurya I casted throwable to HttpException and displayed the user message. The user message in this case was "Unauthenticated. Please login." and I was showing "Invalid Credentials" earlier. So, which one is better? I think "Invalid credentials" is a more clear user message.

therajanmaurya:
fixed in #522

Issue #362, opened 2016-09-27
Path tracker causes app to crash


edcable:
**Summary:** 
When trying to run the path tracker, the app just crashes with the "Unfortunately, MifosX Droid has stopped" error
**Steps to reproduce:** 
From Dashboard, just click Path Tracker

**Expected behavior:** 
Bring up a map that will track location of worker and turn on GPS tracking
**Observed behavior:** 
Just the error
**Device and Android version:** 
Demo Server on HTC One M8 with Android 6.0
**Screenshots:**

sosnickm:
<img width="405" alt="screen shot 2016-10-23 at 2 18 25 am" src="https://cloud.githubusercontent.com/assets/12140915/19624491/0154fc2e-98c7-11e6-977f-92c6f79ec09f.png">

sosnickm:
This is what I get when I click Path Tracker (Android 4.1.1), can you try clicking on it again and seeing if this is just a problem with your version?

edcable:
@therajanmaurya, is this fix included in latest build? My app still crashes when I press start tracking.

therajanmaurya:
fix in #478

PR #398, opened 2016-11-15
Fix : Path tracker causes app to crash #362

reverie-ss:
Fix : Path tracker causes app to crash #362 

Problem : App used to crash on clicking "Start Tracking " button in PathTrackingActivity
Reason : Location permission and request was not handled
Solution : Added run-time location permission request in PatchTrackingActivity


therajanmaurya:
I am closing this issue because this half part of the #399

PR #539, opened 2017-02-22
feat: database encryption using sqlcipher

therajanmaurya:


mifosng-android/src/release/java/com/mifos/api/local/MifosSQLCipherHelper.java
```diff
@@ -0,0 +1,18 @@
...
+public class MifosSQLCipherHelper extends SQLCipherOpenHelper {
+
+    public MifosSQLCipherHelper(DatabaseDefinition databaseDefinition,
+                                DatabaseHelperListener listener) {
+        super(databaseDefinition, listener);
+    }
+
+    @Override
+    protected String getCipherSecret() {
+        return "dbflow-rules";
```
ishan1604: This line of code is the problem, Secret is something that people cannot know. We need to find a way to make it secure.

therajanmaurya: I have two way to implemented.
1. What about if we make the MD5 hash of username and password entered by the user. And save them in the custom data table on the server.
2. Add an extra screen and give the user to enter the database encryption key after the login and save then on server and double encrypt the key and save in Shared Preference So no one can get the key other than admin.

therajanmaurya: @satyan Suggestion

One of the options could be to use user's password entry to encrypt the key on the client. So you could only decrypt with the help of user.

It adds a layer of entropy to the system. However, could mean that we ask the user for password multiple times.. 

Android also supports a hardware backed keychain support. You could potentially explore that as well to store the key.

Having a server backed key is a good option too. Except, in offline case, it'd be difficult to obtain. Do not store in shared preferences, as it's stored in a file on the device in clear. Easily obtained.

PR #559, opened 2017-03-05
fix: LoadMore not working at one case. (#300)

### LINKED ISSUE #300: LoadMore not Working at one case.
LoadMore not woking when we scroll down and load more till the last element and back to top and when we again scrolling to down then LoadMore not working

TypicalGitHubUser:
Fix #300


therajanmaurya:
Good Work 👍

PR #567, opened 2017-03-08
Fix activity not found exception

priyankvex:
The PR contains the patch to prevent the app crash for ActivityNotFoundException and show the user appropriate message.

### DIFF
--- a/mifosng-android/src/main/java/com/mifos/utils/CheckSelfPermissionAndRequest.java
+++ b/mifosng-android/src/main/java/com/mifos/utils/CheckSelfPermissionAndRequest.java
@@ -1,6 +1,7 @@
 package com.mifos.utils;
 
 import android.annotation.TargetApi;
+import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
@@ -11,6 +12,8 @@ import android.provider.Settings;
 import android.support.v4.app.ActivityCompat;
 import android.support.v4.content.ContextCompat;
 import android.support.v7.app.AppCompatActivity;
+import android.util.Log;
+import android.widget.Toast;
 
 import com.mifos.mifosxdroid.R;
 import com.mifos.mifosxdroid.core.MaterialDialog;
@@ -26,6 +29,7 @@ import com.mifos.mifosxdroid.core.MaterialDialog;
  */
 public class CheckSelfPermissionAndRequest {
 
+    private static final String TAG = CheckSelfPermissionAndRequest.class.getName();
 
     /**
      * This Method Check the Permission is granted or not to the App. If the Permission granted,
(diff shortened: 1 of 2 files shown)


mifosng-android/src/main/java/com/mifos/utils/CheckSelfPermissionAndRequest.java
```diff
@@ -128,8 +131,16 @@ public void onClick(DialogInterface dialog, int which) {
                                                 R.string.package_name), activity.getPackageName()
                                                 , null);
                                         intent.setData(uri);
-                                        activity.startActivityForResult(intent,
-                                                Constants.REQUEST_PERMISSION_SETTING);
+                                        try {
```
tarun0: A safe rule of thumb while starting any activity is to first check using `PackageManager` and `Intent.resolveActivity()` whether some activity exists to handle that intent. I think instead of try-catch, you should have used `if (activity exists) {start activity} else {show message}`. What do you say @therajanmaurya ?

It's quite rare though that Settings activity is not found or modded somehow in a way that the usual approach would lead to crash!

mifosng-android/src/main/res/values/strings.xml
```diff
@@ -646,5 +646,5 @@
     <string name="centerList_account_prefix">Account No: %1$s</string>
     <string name="no_staff">no staff</string>
 
-
+    <string name="app_settings_activity_failed">Please open app settings page manually</string>
```
tarun0: `Please open app settings page and grant the permissions manually.` would have been more clear.

mifosng-android/src/main/java/com/mifos/utils/CheckSelfPermissionAndRequest.java
priyankvex: It is crashing on my device (Lenovo k4). I'll change the approach to use the package manager. :+1: 
Thanks.

mifosng-android/src/main/res/values/strings.xml
priyankvex: Will update the message too. :smile:

mifosng-android/src/main/java/com/mifos/utils/CheckSelfPermissionAndRequest.java
```diff
@@ -128,8 +131,16 @@ public void onClick(DialogInterface dialog, int which) {
...
+                                            activity.startActivityForResult(intent,
+                                                    Constants.REQUEST_PERMISSION_SETTING);
+                                        } catch (ActivityNotFoundException exception) {
+                                            Toast.makeText(activity,
+                                                    R.string.app_settings_activity_failed
+                                                    , Toast.LENGTH_SHORT)
+                                                 .show();
+                                            Log.d("MifosApp", "cant find activity with name");
```
devkhan: Using a constant member string as a tag is better.

ishan1604: Good Catch @devkhan 

TAG should be a constant member name and also meaningful! 
MifosApp as a tag is bad choice for debugging purposes :)

priyankvex:
I am creating new PR with the enhancements suggested.

priyankvex:
@tarun0 I have made the changes

Issue #720, opened 2017-07-25
Add sync for surveys


mayank-kgp:
**Summary:** 

Add sync for surveys so that user can see surveys in offline mode.

PR #721, opened 2017-07-25
Sync for Surveys

mayank-kgp:

#720


mayank-kgp:
@therajanmaurya @puneetkohli521 Please review it.
surveysync

therajanmaurya:
@mayank-kgp Please fix the conflict

mayank-kgp:
@therajanmaurya Conflict resolved.

mayank-kgp:
@therajanmaurya Conflict resolved.

mifosng-android/src/main/java/com/mifos/mifosxdroid/online/clientdetails/ClientDetailsFragment.java
```diff
@@ -244,9 +244,9 @@ public void onPrepareOptionsMenu(Menu menu) {
             menu.add(Menu.NONE, MENU_ITEM_DOCUMENTS, Menu.NONE, getString(R.string.documents));
             menu.add(Menu.NONE, MENU_ITEM_UPLOAD_SIGN, Menu.NONE, R.string.upload_sign);
             menu.add(Menu.NONE, MENU_ITEM_IDENTIFIERS, Menu.NONE, getString(R.string.identifiers));
-            menu.add(Menu.NONE, MENU_ITEM_SURVEYS, Menu.NONE, getString(R.string.survey));
             menu.add(Menu.NONE, MENU_ITEM_NOTE, Menu.NONE, getString(R.string.note));
         }
+        menu.add(Menu.NONE, MENU_ITEM_SURVEYS, Menu.NONE, getString(R.string.survey));
```
therajanmaurya: I don't think this line need any change.

mifosng-android/src/main/java/com/mifos/mifosxdroid/online/surveylist/SurveyListPresenter.java
```diff
@@ -35,14 +41,13 @@ public void attachView(SurveyListMvpView mvpView) {
     @Override
     public void detachView() {
         super.detachView();
-        if (mSubscription != null) mSubscription.unsubscribe();
+        if (mSubscriptions != null) mSubscriptions.unsubscribe();
```
therajanmaurya: Use `mSubscriptions.clear()`

mifosng-android/src/main/java/com/mifos/objects/survey/ComponentDatas.java
```diff
@@ -5,16 +5,35 @@
...
+
 /**
  * Created by Rajan Maurya on 28/3/16.
  */
-public class ComponentDatas {
+@Table(database = MifosDatabase.class)
+public class ComponentDatas extends MifosBaseModel implements Parcelable {
+
+    @PrimaryKey
+    int id;
```
therajanmaurya: don't primitive type. use class type.

mifosng-android/src/main/java/com/mifos/objects/survey/QuestionDatas.java
```diff
@@ -6,21 +6,54 @@
...
 import java.util.List;
 
 /**
  * Created by Nasim Banu on 27,January,2016.
  */
-public class QuestionDatas {
+@Table(database = MifosDatabase.class)
+public class QuestionDatas extends MifosBaseModel implements Parcelable {
+
+    @PrimaryKey
```
therajanmaurya: same as above don't use primitive type use class type.

mifosng-android/src/main/java/com/mifos/mifosxdroid/online/clientdetails/ClientDetailsFragment.java
mayank-kgp: @therajanmaurya This line needs to be outside if statement. Because we are enabling offline mode in client list activity. After that when we are navigating to client details activity, none of the menu item is visible because boolean "isClientActive" is false.

therajanmaurya: No, It was good as before because If the client is not active, the user can't do any thing regarding the client. The user should wait till that client will be active.

mayank-kgp: @therajanmaurya Even if the client is active, boolean "isClientActive" will be false. Because for setting it true, an internet connection will be required. But we are disabling connection in ClientListActivity, so by default, this variable will be false.

therajanmaurya: @mayank-kgp App has two modes online and offline, If user has an internet connection then It comes from the server or another wise check in local database and In case the user doesn't sync to the local database then It does not make sense to do any thing with the client.

mayank-kgp: @therajanmaurya Ok. So we will show offline survey only if that client is synced.

therajanmaurya: off course or If the user has internet connection.

therajanmaurya: If you any other use case, tell me. I don't find any use case here.

mayank-kgp:
@therajanmaurya I have made all the changes.

Issue #787, opened 2017-12-24
Log out button not working


AkshGautam:
**Summary:** 
Log out button not working as expected
**Steps to reproduce:** 
Click the log out button from the options menu.
**Expected behavior:** 

App should log out.

**Observed behavior:** 

App not logging out,same screen reappears after entering password.

**Video Link**

[video](https://drive.google.com/file/d/1stqY66hnuECYqA4yxS-ha9XUeWGq1HZP/view?usp=sharing)

If someone else could acknowledge the behaviour shown by the app,I would like to give it a try to debug this. Thanks

AkshGautam:
@therajanmaurya @edcable @dilpreet96  can someone acknowledge about the same behaviour?

tarun0:
It seems that the user can still log in using the Passcode even after clicking Logout button. A better flow would be to make it mandatory for the user to log in using password after clicking on the 'Logout' from the options menu.

@AkshGautam You may work on it.

AkshGautam:
@tarun0 this task is done. I have made the required changes, User now has to enter 4 pin passcode confirming to logout, upon logout he is redirected to the login page for new session.Should I make a PR or anything else need to be done?

tarun0:
Why is the user required to provide the PIN to Logout? 

Yes, you should send PR when you feel you have fixed the issue. Also, add a GIF to show the changes so that it may be easier for the reviewers to have the glimpse of the fix.

You'll be given feedback on the PR itself.

AkshGautam:
@tarun0 That's what i thought there is no need of PIN to logout but prior to working on the logout button the application asked PIN during logout. I was not sure whether to remove that or not.If you feel having asking PIN is not necessary,i would make changes. Please confirm should i remove the PIN before logout option or not? Thanks

tarun0:
What you perceived as a feature is actually a bug. It most probably is the reason why the bug arose; i.e. when one logs out, the home activity is finished which leads to popping the most recent activity (the passcode activity) from the backstack. 

You have to make sure that the Passcode Activity doesn't exist in backstack once it opens the Home screen (Dashboard activity) and that the passcode is removed when one clicks Log Out. That way, when the user will log out, he'll not be shown the passcode activity and he'll be left with no other option than to log in again using username and password.

AkshGautam:
[...]

PR #790, opened 2017-12-30
Fix:Failure of log-out button

AkshGautam:
Log Out Button is working fine!

### DIFF
--- a/mifosng-android/src/main/java/com/mifos/mifosxdroid/core/MifosBaseActivity.java
+++ b/mifosng-android/src/main/java/com/mifos/mifosxdroid/core/MifosBaseActivity.java
@@ -7,6 +7,7 @@ package com.mifos.mifosxdroid.core;
 
 import android.app.ProgressDialog;
 import android.content.Context;
+import android.content.DialogInterface;
 import android.content.Intent;
 import android.support.v4.app.Fragment;
 import android.support.v4.app.FragmentManager;
@@ -38,7 +39,6 @@ public class MifosBaseActivity extends AppCompatActivity implements BaseActivity
     protected Toolbar toolbar;
     private ActivityComponent mActivityComponent;
     private ProgressDialog progress;
-
 
     @Override
     public void setContentView(int layoutResID) {
@@ -131,13 +131,21 @@ public class MifosBaseActivity extends AppCompatActivity implements BaseActivity
 
     @Override
     public void logout() {
-        if (PrefManager.getPassCodeStatus()) {
-            startActivity(new Intent(this, PassCodeActivity.class));
-        } else {
-            PrefManager.clearPrefs();
-            startActivity(new Intent(this, SplashScreenActivity.class));
-        }
-        finish();
+        new MaterialDialog.Builder().init(MifosBaseActivity.this)
+                .setMessage(R.string.dialog_logout)
+                .setPositiveButton(getString(R.string.logout),
+                        new DialogInterface.OnClickListener() {
+                            @Override
+                            public void onClick(DialogInterface dialog, int which) {
+                                PrefManager.clearPrefs();
+                                startActivity(new Intent(MifosBaseActivity.this,
+                                        SplashScreenActivity.class));
+                                finish();
+                            }
+                        })
+                .setNegativeButton(getString(R.string.cancel))
+                .createMaterialDialog()
+                .show();
     }
 
     public void replaceFragment(Fragment fragment, boolean addToBackStack, int containerId) {
--- a/mifosng-android/src/main/java/com/mifos/mifosxdroid/passcode/PassCodeActivity.java
+++ b/mifosng-android/src/main/java/com/mifos/mifosxdroid/passcode/PassCodeActivity.java
@@ -97,6 +97,7 @@ public class PassCodeActivity extends MifosBaseActivity implements PassCodeView.
         if (isPassCodeLengthCorrect()) {
             if (isPassCodeCorrect) {
                 startDashBoardActivity();
+                finish();
             } else {
                 counter++;
                 passCodeView.clearPasscodeField();
--- a/mifosng-android/src/main/res/values/strings.xml
+++ b/mifosng-android/src/main/res/values/strings.xml
@@ -186,6 +186,7 @@
     <string name="disburse_loan">Disburse Loan</string>
     <string name="logging_in">Logging In</string>
     <string name="back_again">Please press back again to exit</string>
+    <string name="dialog_logout">Are you sure you want to logout?</string>
     <!-- General Terms End Here -->
 
 


AkshGautam:
@tarun0 Please review when you get time :)

tarun0:
- Squash the commits into a single commit
- Follow commit style guideline (find it on the wiki of the project, here at Github)
- Don't add unnecessary blank lines
- Add a GIF in the PR as a comment which demonstrates the changes you've added

mifosng-android/src/main/java/com/mifos/mifosxdroid/core/MifosBaseActivity.java
```diff
@@ -5,6 +5,7 @@
 
 package com.mifos.mifosxdroid.core;
 
+
```
tarun0: What I meant by unnecessary blank lines were these - added by you and not those which Android Studio automatically inserts to distinguish imports from different packages. Please keep the blank lines between the imports as they were.

mifosng-android/src/main/java/com/mifos/mifosxdroid/passcode/PassCodeActivity.java
```diff
@@ -67,6 +65,7 @@ protected void onCreate(Bundle savedInstanceState) {
             //enabling passCodeListener only when user has already setup PassCode
             passCodeView.setPassCodeListener(this);
         }
+
```
tarun0: Remove all the changes in this file except the `finish()` statement.

tarun0:
You may find a tutorial for `Squashing` the commits here: https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

I'd recommend squashing the commits (even if there is just single commit) once the conversation has started on any PR because amending the commit will lead to deletion of the comments on the PR.

mifosng-android/src/main/java/com/mifos/mifosxdroid/core/MifosBaseActivity.java
AkshGautam: done!

mifosng-android/src/main/java/com/mifos/mifosxdroid/passcode/PassCodeActivity.java
AkshGautam: done!

AkshGautam:
@tarun0 i hope, it satisfies every requirement.I got stuck in some git error. I was not able to push my changes.If anything else is required,i would make the necessary changes.Thanks
Here is the gif

AkshGautam:
@tarun0 After exploring a lot with git rebase -i HEAD~X ,i hope i have finally squashed all commits into one. Don't know why i was not able to push though. I had to use git push -f. Thanks for your patience.

mifosng-android/src/main/java/com/mifos/mifosxdroid/passcode/PassCodeActivity.java
```diff
@@ -97,12 +98,14 @@ public void passCodeEntered(String passcode) {
...
             if (isPassCodeCorrect) {
                 startDashBoardActivity();
+                finish();
             } else {
                 counter++;
                 passCodeView.clearPasscodeField();
                 Toaster.show(clRootview, R.string.incorrect_passcode);
             }
         }
+
```
tarun0: You haven't added the screenshot/GIF of the changes yet... Add it so that the reviewers (or other org members) don't have to build the app from your repo to observe the changes.

AkshGautam:
logout

mifosng-android/src/main/java/com/mifos/mifosxdroid/passcode/PassCodeActivity.java
AkshGautam: Done. I am sorry for carelessness.I will be cautious from next time.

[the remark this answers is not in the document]
AkshGautam: I removed the blanked line as well. Would take care of it from now onwards.Sorry for the mistake

mifosng-android/src/main/java/com/mifos/mifosxdroid/core/MifosBaseActivity.java
```diff
@@ -131,13 +132,21 @@ public void hideKeyboard(View view) {
...
-        } else {
-            PrefManager.clearPrefs();
-            startActivity(new Intent(this, SplashScreenActivity.class));
-        }
-        finish();
+        AlertDialog.Builder builder = new AlertDialog.Builder(this);
+        builder.setMessage(R.string.logout_message)
+                .setPositiveButton(R.string.action_yes, new DialogInterface.OnClickListener() {
+                    public void onClick(DialogInterface dialog, int id) {
+                        PrefManager.clearToken();
```
tarun0: Remove all the sharedpreferences here instead of using `clearToken()` method.

When one logs out, the application should clear all the saved 'SharedPreference values'. `clearToken()` will only save a blank passcode value and all the remaining sharedpreferences will remain intact. You can save the `Instance Domain` to make it convenient for the other users to login but it is already handled in `LoginActivity`.

tarun0: Also, could you double check whether this issue of white text is present on other devices as well? My device is Motorola G5S Plus, Stock Nougat (7.1.1) - API 25.

pr

AkshGautam: @tarun0 I thought of clearing all the shared preferences at first but then I realized that if a user logs out and the logs in again his identity is lost and he will be a new user to the application making him to set the 4-digit passcode on subsequent logins.I will clear all the preferences for now and would start working on the problem of preventing the reset of password. I have already created an issue for that in mifos mobile as well.

tarun0: ` if a user logs out and the logs in again his identity is lost and he will be a new user to the application making him to set the 4-digit passcode on subsequent logins` 
Isn't is exactly how it should be? Why do you think passcode should be retained after logout?

AkshGautam: @tarun0 Let's take an example of Paytm,The user creates a 4 digit PIN which is asked before any operation involving banking.The 4 digit passcode can be used instead of a longer PIN to ensure authenticity. Once the user logs-out the application(does not mean he wont use the app again,so why need to reset the password  ) the password is still saved since the user is expected to use the app again.That's my opinion.If you feel this is irrelavent i would amend the code.

tarun0: Your point is not irrelevant or wrong. However, you are mistaking the passcode used in the Mifos App with the PIN used in banking apps. 

 - While the PIN is an added security measure in the baking apps, the passcode in Mifos is a tad inclined towards convenience as compared to security. It provides a way for the user to log into the app with a 4 digit passcode saving him from entering username and password every time he opens the app. Our intention to add such a feature is that unauthorized user (who has physical access to the phone) shouldn't use be allowed to use the app when the authenticated user has already exited it (in the future versions, it's supposed to evolve into a feature that would ask for passcode every time before the user performs any writable action, e.g. apply for a loan).

 - The PIN used in banking apps is stored on the server side to verify the authenticity whereas, in our app, the Passcode is saved in SharedPreference (hashed form) only and the verification is done locally, without any kind of intervention from the server.

Still, there's some flaw in your proposition. If the user logs in from device 1 and saves the passcode and then logs out, assuming that the passcode still refers that particular user, what will happen if he logs in from device 2? How will device 1 get information that a new passcode has been generated on some other device? The solution to this intervention of the server.

Your proposition was something in the pipeline but it might affect offline functionality and would add to more network calls. As I'm not the right person to confirm or reject any possibility whether this feature will be added, I can only tell you that it's not present currently.

I hope it provides some clarification. :)

AkshGautam: @tarun0 Alright,I understood the functionality of the passcode now. Moreover retaining the passcode would be consistent only if the application is hosted on a universally available database on cloud. I have made the changes by clearing the preferences on logout.Thanks

AkshGautam: @tarun0 I would check this in some other device as well and post the update.

AkshGautam: @tarun0  I have checked the application on Android 6.0.1,API 23 and on Android 5.0.2 API 21 and I did not find the white text problem.

AkshGautam:
@tarun0 Could you please review and suggest any improvements so it could be merged.Thanks

therajanmaurya:
@AkshGautam  Good work, Can you please make little changes, We already have Material Dialog Please make changes in PR and use this https://github.com/openMF/android-client/blob/master/mifosng-android/src/main/java/com/mifos/mifosxdroid/core/MaterialDialog.java

AkshGautam:
@therajanmaurya sure! I would make the changes.

AkshGautam:
@therajanmaurya @tarun0  done the changes by adding material dialog as well.

AkshGautam:
@therajanmaurya Issue Number #787

Issue #823, opened 2018-02-03
Improper Resources inflation in Create Client activity


AkshGautam:
**Summary:** 

The spinners for creating new client activity are not working properly.

**Steps to reproduce:** 

1)Run the app
2)Create Client

**Expected behavior:** 

The application should inflate all the resources in user friendly and correct way.

**Observed behavior:** 

The gender spinner,Client,Client Classification was not inflated for a long time and then it abruptly got inflated.

**Device and Android version:** 

Android 5.0.2,API 21

**Screenshots:** 

![20180203_150501](https://user-images.githubusercontent.com/19329788/35765953-a3129172-08f4-11e8-9d0b-e301d0dd9a81.gif)

AkshGautam:
@therajanmaurya @tarun0  I am working on this.

PR #825, opened 2018-02-04
Fix-Issue#823 :Improper Layout Inflation

AkshGautam:
Fix: #823


AkshGautam:
@therajanmaurya @tarun0 I have fixed the issue but this class has to be made more robust. There are different threads for loading offices, loading staff in offices, loading client in the template. The threads are only run when the scheduler policy gives them priority and thus the thread which loads client in template gets executed later due to which the array adapter data is added a lot later comparative to other threads which finish their execution a lot before. This asynchronous execution could be made more robust and the overhead of context switching could be reduced after discussion.

PR #1006, opened 2018-12-09
Fix #1005: Update okHttp version

### LINKED ISSUE #1005: HTTP/2 HPACK header table size update not understood by the client
**Summary:** 

MifosXdroid or one of the dependencies does not understand specific HTTP/2 protocol feature.

**Steps to reproduce:** 

1. Install MifosX on Apache Tomcat
2. Set nginx version 1.13.6 or newer as reverse proxy for the Tomcat
3. Enable HTTP/2 and TLS on the nginx reverse proxy
4. Attempt to connect to the instance using MifosXdroid client

**Expected behaviour:** 

Client can connect (ie. either successfully log in or display a message about wrong credentials)

**Observed behaviour:** 

Clien't can't connect to the instance. It receives data, but nothing is shown to the user and application remains on login screen.

This may be a bit low-level, so bear with me... MifosXdroid v1.0.1 does not understand properly a web server response containing **HTTP/2 HPACK header table size update** information. We used to have Mifos X behind a nginx **1.12.2** reverse proxy with TLS and HTTP/2 enabled. After we have upgraded nginx to **1.14.1**, MifosXdroid was no longer able to connect to the server. No error message was displayed on either client or server side. In fact no messages were displayed at all, as if the response was discarded by the client. Server logs indicated that the requests and responses are accepted and returned as expected.

Finally, I have traced the connections using tcpdump and wireshark and decrypted them to see the differences between the old and the new setup. The only meaningful difference was, that nginx since version 1.13.6 (commit https://github.com/nginx/nginx/commit/63f5d46f58e537745987e9d68bc65b03830238a7) adds HTTP/2 0-byte HPACK table size update information to inform that it doesn't use HPACK's dynamic table (which is in accordance with [RFC7540](https://tools.ietf.org/html/rfc7540) section 6.5.2 and [RFC 7541](https://tools.ietf.org/html/rfc7541)). When I stripped this information using intercepting transparent proxy (mitmproxy), the client was able to connect again.

This is presumably not a problem of the codebase but of one of the third-party dependencies (most likely *okhttp*). In order to fix the problem, this dependency either needs to be updated or, if the bug is present even in the newest version, reported in the appropriate upstream. (I have zero experience with Android development to do so myself, I'm a sysadmin)

HTC 10, stock ROM, Android 8.0.0. Tested also on bunch of others.

**Screenshots:** 

hpack

trendspotter:
Fixes #1005 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ ext {
     runnerVersion = '0.4.1'
     rulesVersion = '0.4.1'
     retrofitVersionLatest = '2.1.0'
-    okHttp3Version = '3.3.1'
+    okHttp3Version = '3.5.0'
     flipTableVersion = '1.0.1'
     butterKnifeVersion = '8.0.1'
     mifosPasscodeVersion = '0.3.0'


droidchef:
@trendspotter Thank you so much for the in depth investigation of the problem and the fix. Highly appreciate your effort in supporting the project through your valuable contribution.

PR #1169, opened 2019-08-10
Replaced support libraries with AndroidX and made some necessary chan…

moksh-mahajan:
…ges for the upgrade


mifosng-android/src/main/java/com/mifos/mifosxdroid/core/MifosBaseActivity.java
```diff
@@ -122,9 +122,12 @@ public void hideProgress() {
 
     public void hideKeyboard(View view) {
         InputMethodManager inputManager = (InputMethodManager) this
-                 .getSystemService(Context.INPUT_METHOD_SERVICE);
-        inputManager.hideSoftInputFromWindow(getCurrentFocus().getWindowToken(), InputMethodManager
-                 .RESULT_UNCHANGED_SHOWN);
+                .getSystemService(Context.INPUT_METHOD_SERVICE);
+        View focusedView = this.getCurrentFocus();
+        if (focusedView != null) {
```
tarun0: 1. This change is not related to AndroidX porting. So, it shouldn't have been here. 
2. What if the view is null?

moksh-mahajan: Actually it's a UI bug. I will create a separate issue for that.

PR #1290, opened 2020-02-03
Fix: Error message in Snackbar for withdrawal in saving accounts.

### LINKED ISSUE #1132: Unable to view Error Message when keyboard open
**Summary:** 

In Client Savings Account Transaction Activity, if the keyboard is open and a error message pops up, it appears behind the keyboard and isn't visible to the user. This can cause confusion to the user as he is stuck on the same page with no message informing him why.

**Steps to reproduce:** 
1. Go to the Savings Account of any Client and try to withdraw more than available in the account.
2. Make sure to keep the keyboard open and then press Review Transaction and then press Process Transaction.

**Expected behaviour:** 
An error message should pop up informing the user that the transaction failed.

**Observed behaviour:** 
The snackbar appears behind the keyboard and thus can't be seen. Only when you dismiss the keyboard is the Snackbar seen. This may create confusion for users who don't dismiss the keyboard to see it as it will appear as if nothing has happened.

Redmi Note 4 Lineage OS 15.1

**Screenshots:** 
Here after I click Process Transaction, it just appears as though the page has just refreshed and nothing has occured. However only on dismissing the keyboard do I see the Snackbar "Transaction Failed"  at the bottom. 
keyboard (1)

robustTechie:
Fixes #1132 

resolved27


Aditya-gupta99:
We are closing this PR as it causes a branch conflict since the Java file has now been migrated to kotlin

PR #1297, opened 2020-02-10
Fix: Modified search result with spinner menu All

### LINKED ISSUE #1296: "All" option in Spinner not updating the search result
**Summary:** 

If Spinner in Run Reports is changed to All then it does not show all results.  It only shows the previous search result.

**Steps to reproduce:** 

open nav drawer> Run reports> Change the spinner to All

**Expected behaviour:** 

After changing the spinner value to All it should show everything in the search result.

**Observed behaviour:** 

If Spinner in Run Reports is changed to All then it does not show all results.  It only shows the previous search result.

Android Version 9.0
Realme 5

**Screenshots:** 

isuue29

iamsh4shank:
Fixes #1296

Fixed the alignment of All menu in spinner also added a switch case for All. But the problem is, it doesn't have any backend for fetching All the runreports at a same time. So if the All feature is added in the backend this feature will show all the runreports.

resolved30


Aditya-gupta99:
We are closing this PR as it causes a branch conflict since the Java file has now been migrated to kotlin

PR #1324, opened 2020-02-21
Fix #1281: Null value displayed in Client Activity toolbar

### LINKED ISSUE #1281: Null value displayed in Client Activity toolbar
**Summary:** 

Null value is displayed in Client Activity toolbar if last name of the client does not exist

**Steps to reproduce:** 

1. Nav Bar -> Clients
2. Click on a client for which surname does not exist
3. Null display on the toolbar

**Expected behaviour:** 

In this case first name should be displayed

**Observed behaviour:** 

Instead, null value is displayed

Android 10
OnePlus 6
Stock ROM

**Screenshots:** 

null client name

Grandolf49:
Fixes #1281 

fix - null value in client activity

### DIFF
--- a/mifosng-android/src/main/res/layout/activity_login.xml
+++ b/mifosng-android/src/main/res/layout/activity_login.xml
@@ -14,6 +14,7 @@
         >
 
         <LinearLayout style="@style/LinearLayout.Base"
+            android:animateLayoutChanges="true"
             >
 
             <ImageView


mifosng-android/src/main/java/com/mifos/mifosxdroid/online/clientdetails/ClientDetailsFragment.java
```diff
@@ -503,7 +504,11 @@ public void showProgressbar(boolean show) {
     @Override
     public void showClientInformation(final Client client) {
         if (client != null) {
-            setToolbarTitle(getString(R.string.client) + " - " + client.getLastname());
+            String lastname = client.getLastname();
+            if (lastname == null) {
+                lastname = client.getDisplayName();
```
droidchef: What if the `client.DisplayName()` also returns null?

Grandolf49: We have the following options for it.
1. `client.getLastName()`
2. `client.getDisplayName()`
3. `client.getFirstName()`
4. `client.getMiddleName()`

I think we can check which name exists in this particular order.
if(we find a name which is not null):
    set that name in the toolbar
else
    don't set anything, just keep **ClientActivity** as the toolbar title

@droidchef What are your thoughts on this?

Grandolf49: @luckyman20 @tarun0 @droidchef @therajanmaurya After researching more about this and the web client, there are 2 options for the client's legal form:
1. If Legal Form is **Person**
- Client will have the First Name and Last Name field. In this case `client.getLastName()` and `client.getFirstName()` will never return null
- Here `client.getDisplayName()` will return a string in the form `firstname + middlename + lastname`
 
2. If Legal Form is **Entity**
- Client will only have the field Name and nothing else. In this case `client.getFirstName()` and `client.getLastName()` both will return null
- Here `client.getDisplayName()` will return the correct name

Thus, in both the cases, `client.getDisplayName()` will never return null. So I think we should use that. Please let me know if I can go ahead and implement this

Grandolf49: I have created a new PR for the same. Please review it #1438

PR #1387, opened 2020-03-04
fix #1386: No progress dialog for Savings Account Activation

### LINKED ISSUE #1386: No progress dialog for Savings Account Activation
**Summary:** 

Progress dialog does not appear if user tries to activate savings account

**Steps to reproduce:** 

1. Search for **Mifos** and apply **Savings Account** Filter.
2. Select any savings account
3. Select Activate Savings
4. Click save. There is no progress dialog after user clicks save.

**Expected behaviour:** 

There should be a progress dialog after the user presses save button.

Android 10
OnePlus 6
Stock ROM

**Screenshots:** 

no progress bar

Grandolf49:
Fixes #1386 

**Screenshots**

fix - no progress bar


robustTechie:
I think the ProgressBar will be good when that save button is working like saving the accounts. As now nothing is happening instead of a toast message so it will be better if now we don't use progressBar.

Grandolf49:
@robustTechie As the api requires some time to fetch results, till that time nothing is happening in the app which is not good from UX pov. So I thought it is better to provide the progress bar instead of leaving the user wonder what's happening in the app

robustTechie:
Ok

PR #1388, opened 2020-03-05
Fix: Search result modify

### LINKED ISSUE #1379: Search result not update
**Summary:** 

If we change the filter and no such search result is found then it will show the previous search result.

**Steps to reproduce:** 

Search any name in the dashboard and then change the filter such that the filter doesn't have search results.

**Expected behaviour:** 

If no result is found then it should clean the search result.

**Observed behaviour:** 

If we change the filter and no such search result is found then it will show the previous search result.

Android Version: 9.0
Realme 5

**Screenshots:** 

issue39-1

robustTechie:
Fixes #1379

resolved37

### DIFF
--- a/mifosng-android/src/main/java/com/mifos/mifosxdroid/online/search/SearchFragment.java
+++ b/mifosng-android/src/main/java/com/mifos/mifosxdroid/online/search/SearchFragment.java
@@ -100,12 +100,6 @@ public class SearchFragment extends MifosBaseFragment implements SearchMvpView,
         sp_search.setAdapter(searchOptionsAdapter);
         sp_search.setOnItemSelectedListener(this);
         et_search.requestFocus();
-        layoutManager = new LinearLayoutManager(getActivity());
-        layoutManager.setOrientation(LinearLayoutManager.VERTICAL);
-        rv_search.setLayoutManager(layoutManager);
-        rv_search.addOnItemTouchListener(new RecyclerItemClickListener(getActivity(), this));
-        rv_search.setHasFixedSize(true);
-        rv_search.setAdapter(searchAdapter);
     }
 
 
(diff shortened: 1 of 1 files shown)


mifosng-android/src/main/java/com/mifos/mifosxdroid/online/search/SearchFragment.java
```diff
@@ -124,6 +118,12 @@ public void onClickSearch() {
 
     @Override
     public void showSearchedResources(List<SearchedEntity> searchedEntities) {
+        layoutManager = new LinearLayoutManager(getActivity());
+        layoutManager.setOrientation(LinearLayoutManager.VERTICAL);
+        rv_search.setLayoutManager(layoutManager);
+        rv_search.addOnItemTouchListener(new RecyclerItemClickListener(getActivity(), this));
+        rv_search.setHasFixedSize(true);
+        rv_search.setAdapter(searchAdapter);
         searchAdapter.setSearchResults(searchedEntities);
```
jawidMuhammadi: Please revert all these changes, it is not a good practice from memory optimization and memory allocation perspective because creating layout manager object,  setting itmeTouchListener, size, and adapter to search view every time when we make search is a heavier task and needs memory allocation for each that is why the previous place is better, there, it needs to be initialized once during the creation of fragment only.

mifosng-android/src/main/java/com/mifos/mifosxdroid/online/search/SearchFragment.java
```diff
@@ -133,6 +133,7 @@ public void showSearchedResources(List<SearchedEntity> searchedEntities) {
     public void showNoResultFound() {
         showAlertDialog(getString(R.string.dialog_message),
                 getString(R.string.no_search_result_found));
+        rv_search.setAdapter(null);
```
jawidMuhammadi: setting adapter null in searchView sometimes leads to unexpected crashes. Instead of this, you can set resultSet array of adapter to an empty list like this `searchAdapter.setSearchResults(Collections.<SearchedEntity>emptyList());` .

PR #1481, opened 2020-04-07
Fix #1480: Deleted duplicate branches in Switch statements

### LINKED ISSUE #1480: Duplicate branches in Switch statements
**Summary:** 

This issue focuses on the `switch` statements containing the same code in different branches. The redundant cases can be removed as they are of no use and cause code redundancy.   These types of switch cases are present in `SyncCenterPayloadsFragment` and `SyncPayloadsAdapter`.

`Example:`
image

**Expected behaviour:** 

Duplicate branches in switch statements must be removed.

**Observed behaviour:** 

Duplicate branches in Switch statements.

RajaVamsi11:
Fixes #1480

### DIFF
--- a/mifosng-android/src/main/java/com/mifos/mifosxdroid/adapters/SyncPayloadsAdapter.java
+++ b/mifosng-android/src/main/java/com/mifos/mifosxdroid/adapters/SyncPayloadsAdapter.java
@@ -59,9 +59,6 @@ public class SyncPayloadsAdapter extends RecyclerView.Adapter<RecyclerView.ViewH
             ((ViewHolder) holder).tv_activation_date.setText(clientPayload.getActivationDate());
 
             switch (clientPayload.getGenderId()) {
-                case 22 :
-                    ((ViewHolder) holder).tv_gender.setText("Male");
-                    break;
                 case 24 :
                     ((ViewHolder) holder).tv_gender.setText("Female");
                     break;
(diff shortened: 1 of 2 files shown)


mifosng-android/src/main/java/com/mifos/mifosxdroid/adapters/SyncPayloadsAdapter.java
```diff
@@ -59,9 +59,6 @@ public void onBindViewHolder(RecyclerView.ViewHolder holder, int position) {
             ((ViewHolder) holder).tv_activation_date.setText(clientPayload.getActivationDate());
 
             switch (clientPayload.getGenderId()) {
-                case 22 :
-                    ((ViewHolder) holder).tv_gender.setText("Male");
```
jawidMuhammadi, line 63: Why we are deleting this line of code?

RajaVamsi11: To remove redundancy. @jawidMuhammadi The piece of code, in case 22, is the same as in default. There is no need for extra cases as it will be implemented in the default case if there is no particular case.

RajaVamsi11: The functionality will remain the same even if we delete the above case. That's why we are deleting the above line of code.

jawidMuhammadi: Actually, here, even the `switch case` is not a good place to use instead you can use simply `if-else` statement. Although it is always recommended to handle all cases ( `else` ) in a recyclerView adapter; because whenever we scroll, recycleView recycles the used Items of the screen for new items sometimes it can provide unwanted results or behavior if we don't handle all cases due to catch in memory.

PR #1500, opened 2020-04-13
Replaced addAll() method call with parameterized constructor call

### LINKED ISSUE #1499: addAll() method call can be replaced with parameterized constructor call
**Summary:** 

The method calls such as **Collection.addAll()** and **Map.putAll()** after instantiation of a collection using a constructor call without arguments. 

Such constructs can be replaced with a single call to a parameterized constructor which simplifies code. Also for some collections, the replacement might be more performant.
For example:
 ```
 Set<String> set = new HashSet<>();
  set.addAll(Arrays.asList("alpha", "beta", "gamma"));
```
can be replaced with:

`Set<String> set = new HashSet<>(Arrays.asList("alpha", "beta", "gamma"));`

### PR DESCRIPTION
**Pull request by RajaVamsi11**
Fixes #1499

### DIFF
--- a/mifosng-android/src/main/java/com/mifos/mifosxdroid/online/runreports/reportcategory/ReportCategoryPresenter.java
+++ b/mifosng-android/src/main/java/com/mifos/mifosxdroid/online/runreports/reportcategory/ReportCategoryPresenter.java
@@ -97,8 +97,7 @@ public class ReportCategoryPresenter extends BasePresenter<ReportCategoryMvpView
         for (ClientReportTypeItem item : list) {
             map.put(item.getReportId(), item);
         }
-        ArrayList<ClientReportTypeItem> uniques = new ArrayList<>();
-        uniques.addAll(map.values());
+        ArrayList<ClientReportTypeItem> uniques = new ArrayList<>(map.values());
         return uniques;
     }
 }

### CODE REVIEW

File: mifosng-android/src/main/java/com/mifos/mifosxdroid/online/runreports/reportcategory/ReportCategoryPresenter.java [hunk 1]
```diff
@@ -97,8 +97,7 @@ public void onNext(List<ClientReportTypeItem> clientReportTypeItems) {
         for (ClientReportTypeItem item : list) {
             map.put(item.getReportId(), item);
         }
-        ArrayList<ClientReportTypeItem> uniques = new ArrayList<>();
-        uniques.addAll(map.values());
+        ArrayList<ClientReportTypeItem> uniques = new ArrayList<>(map.values());
```
jawidMuhammadi, line 100: Why and what makes it different from the existing code? is there any benefit to it?

RajaVamsi11: Just a refactor.

jawidMuhammadi: Nice, but I think there are hundreds of line code like this in the entire project if we raise PR for each one of them separately and even if there is no benefit of it or without any specific reason, don't you think the number of PRs increases too much and it will be difficult and time-consuming to maintainers or mentors to merge them or because of this they might miss or delay to merge the important PRs?

RajaVamsi11: Ok Jawid, I will keep that in mind in future. Thankyou :)

PR #1612, opened 2020-12-09
Fixes #1610 InstanceURl edit preference added in settings

### LINKED ISSUE #1610: No option for user to change instance url when signed in
**Summary:** 

Once any user has signed in the app. He/she cannot change the instance url. Although there should be a way to do so without resigning and reconfiguring the network setting.

**Expected behaviour:** 

In settings there should be option for changing the instance url.

**Observed behaviour:** 

No option in the app for changing instance url after signing in.

* ANY

danishjamal104:
Fixes #1610  

ss-2 | ss-1

### DIFF
--- a/mifosng-android/src/main/java/com/mifos/mifosxdroid/SettingsFragment.kt
+++ b/mifosng-android/src/main/java/com/mifos/mifosxdroid/SettingsFragment.kt
@@ -2,28 +2,23 @@ package com.mifos.mifosxdroid
 
 import android.content.Intent
 import android.content.SharedPreferences
+import android.os.Build
 import android.os.Bundle
-import android.preference.ListPreference
+import android.preference.*
 import android.preference.Preference.OnPreferenceChangeListener
-import android.preference.PreferenceFragment
-import android.preference.SwitchPreference
 import android.widget.Toast
 import com.mifos.mifosxdroid.dialogfragments.syncsurveysdialog.SyncSurveysDialogFragment
-import com.mifos.utils.FragmentConstants
-import com.mifos.utils.LanguageHelper
-import com.mifos.utils.ThemeHelper
-import android.preference.Preference
-import android.preference.PreferenceScreen
-
+import com.mifos.mifosxdroid.online.DashboardActivity
 import com.mifos.mifosxdroid.passcode.PassCodeActivity
 import com.mifos.mobile.passcode.utils.PasscodePreferencesHelper
-import com.mifos.utils.Constants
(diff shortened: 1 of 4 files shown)


danishjamal104:
@robustTechie please review the changes. thanks :)

robustTechie:
Hey, @danishjamal104 how about adding the passcode verification just before changing the Instance URL ?

robustTechie (commented):
Hey, @danishjamal104 how about adding the passcode verification just before changing the Instance URL ?

danishjamal104:
> Hey, @danishjamal104 how about adding the passcode verification just before changing the Instance URL ?

I am not able to get the actual idea of how things are related. Like what's the difference between domain and instance and the changes that I made only updates the sharedPref using Prefmanager as you can. see in changes. I wasn't clear about it that's why I didn't applied passcode functionality.
So please help me with that.

robustTechie:
Yeah, so you can something like if the user clicks on change instance thing then it will ask the user to enter the passcode then allows to change the URL

iamsh4shank (approved):

iamsh4shank:
@danishjamal104 any update on this?

danishjamal104:
> @danishjamal104 any update on this?

I think passcode isn't required while changing the instance URL, because its more of a administration thing, and while starting the app also user has option to change the domain, port, and tenant which is before setting password.

mifosng-android/src/main/java/com/mifos/mifosxdroid/SettingsFragment.kt
```diff
@@ -54,6 +49,28 @@ class SettingsFragment : PreferenceFragment(), SharedPreferences.OnSharedPrefere
             true
         }
 
+        mInstanceUrlPref = findPreference(
+                getString(R.string.hint_instance_url)) as EditTextPreference
+        val instanceUrl = PrefManager.getInstanceUrl()
+        mInstanceUrlPref!!.text = instanceUrl
+        mInstanceUrlPref!!.isSelectable = true
+        mInstanceUrlPref!!.dialogTitle = "Edit Instance Url"
```
iamsh4shank, line 57: It would be great if you could change this hardcoded string into string resources :)

iamsh4shank (changes requested):

iamsh4shank (changes requested):
Hwy @danishjamal104 can you fix the requested changes as well as merge conflicts so that we could merge it soon :)

iamsh4shank (approved):

PR #2223, opened 2024-11-07
Clean up ->Center module

Darkeye14:
Fixes #Issue_Number

Recording:

https://github.com/user-attachments/assets/40ae706b-58dc-4e83-b116-3045b4d131aa

- [ ] Apply the `MifosStyle.xml` style template to your code in Android Studio.

- [ ] Run the unit tests with `./gradlew check` to make sure you didn't break anything


mifosng-android/src/main/java/com/mifos/mifosxdroid/dialogfragments/synccenterdialog/SyncCentersDialogFragment.kt
```diff
@@ -1,62 +1,62 @@
...
+//        centers?.let {
+//            viewModel.setCentersList(centersList = centers)
+//        }
+//        super.onCreate(savedInstanceState)
+//    }
+//
+//    override fun onCreateView(
+//        inflater: LayoutInflater, container: ViewGroup?,
+//        savedInstanceState: Bundle?
+//    ): View {
```
itsPronay, line 33: I would like to mention a few things

1. We need to get rid of fragments instead of commenting them out. As for the syncCentersDialog dialog fragment, we need to check the old code and see where it used to be called. Then, we have to implement it in compose code. 
2. We follow single commit structure here, so please squash your commits.

PR #2312, opened 2025-02-14
refactor: [database/charge] dbflow to room

itsPronay:
Fixes - https://mifosforge.jira.com/browse/MIFOSAC-392


core/data/src/main/java/com/mifos/core/data/pagingSource/ClientChargesPagingSource.kt
```diff
@@ -55,22 +50,16 @@ class ClientChargesPagingSource(
...
-                .subscribeOn(Schedulers.io())
-                .subscribe(object : Subscriber<Page<Charges>>() {
-                    override fun onCompleted() {
-                    }
-
-                    override fun onError(exception: Throwable) {
-                        continuation.resumeWithException(exception)
-                    }
+        return try {
+            val page = dataManagerCharge.getClientCharges(
```
therajanmaurya: See documentation, How to use coroutines in PagingSource

core/database/src/main/java/com/mifos/room/dao/ChargeDao.kt
```diff
@@ -0,0 +1,30 @@
...
+import kotlinx.coroutines.flow.Flow
+
+/**
+ * Created by Pronay Sarker on 14/02/2025 (3:32 PM)
+ */
+@Dao
+interface ChargeDao {
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insertCharges(charge: List<Charges>)
```
therajanmaurya: instead of `charge: List<Charges>` `vararg charges: Charges`

core/network/src/main/java/com/mifos/core/network/services/ChargeService.kt
```diff
@@ -37,11 +39,11 @@ interface ChargeService {
     suspend fun getAllChargeV3(@Path("loanId") loanId: Int): ResponseBody
 
     @GET(APIEndPoint.CLIENTS + "/{clientId}/charges")
-    fun getListOfCharges(
+    suspend fun getListOfCharges(
         @Path("clientId") clientId: Int,
         @Query("offset") offset: Int,
         @Query("limit") limit: Int,
-    ): Observable<Page<Charges>>
+    ): Page<Charges>
```
niyajali: This could be Flow<T>

core/database/src/main/java/com/mifos/room/helper/ChargeDaoHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+        clientId: Int,
+    ) {
+        val updatedCharges = chargesPage.pageItems.map { charges ->
+            val dateParts = charges.dueDate.orEmpty().split("-").mapNotNull { it.toIntOrNull() }
+
+            val clientDate = if (dateParts.size == 3) {
+                charges.id?.toLong()?.let {
+                    ClientDate(
+                        0,
+                        it,
```
niyajali: Always try to pass parameters names for better Readability

niyajali (changes requested):

core/database/src/main/java/com/mifos/room/helper/ChargeDaoHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * This method Retrieve the Charges from Charges_Table and set the Charges Due date after
+     * loading the Charge due date from the ChargeDate_table as reference with charge Id.
+     *
+     * @param clientId Client ID
+     * @return Page of Charges
+     */
+    fun readClientCharges(clientId: Int): Flow<Page<Charges>> {
+        return chargeDao.getClientCharges(clientId).map { chargesList ->
```
niyajali: Inject ioDispatcher in this class and use flowOn(ioDispatcher) oparator, for all methods which returns as Flow<T>, do this in other classes too

And this function implementation looks messy to me, copy this code and chat on Deepseek, Chatgpt etc and ask them to optimize/improve this codebase, nowadays we've many free AI tools consider to use those for better usability.

niyajali (changes requested):

core/network/src/main/java/com/mifos/core/network/services/ChargeService.kt
itsPronay: @niyajali You mean this right? Retrofit does not support flow
```
 fun getListOfCharges(
        @Path("clientId") clientId: Int,
        @Query("offset") offset: Int,
        @Query("limit") limit: Int,
    ): Flow<Page<Charges>>
```

core/database/src/main/java/com/mifos/room/di/DaoModule.kt
```diff
@@ -41,4 +42,9 @@ object DaoModule {
     fun providesStaffDao(database: MifosDatabase): StaffDao {
         return database.staffDao()
     }
+
+    @Provides
+    fun providesClientDao(database: MifosDatabase): ChargeDao {
```
HekmatullahAmin, line 67: @itsPronay I think it would be a good idea to add @Singleton to the providesClientDao() method. Since Room is designed to work with a single shared database connection, using @Singleton ensures that all components use the same ChargeDao instance.

core/data/src/main/java/com/mifos/core/data/pagingSource/ClientChargesPagingSource.kt
```diff
@@ -55,22 +50,16 @@ class ClientChargesPagingSource(
...
+                limit = 10,
+            ).first()
 
-                    override fun onNext(page: Page<Charges>) {
-                        continuation.resume(Pair(page.pageItems, page.totalFilteredRecords))
-                    }
-                })
+            Pair(page.pageItems, page.totalFilteredRecords)
+        } catch (exception: Throwable) {
+            throw exception
```
HekmatullahAmin: I suggest wrapping the throw statement in a custom exception. It will help us to easily trace errors during debugging.
somehting like: 
throw DatabaseFetchException("Failed to fetch client charges", exception)

HekmatullahAmin:
@niyajali Based on what I mentioned earlier, I didn’t make changes to the ViewModel because the focus right now is on completing the migration, as discussed in the stand-up. However, if you still feel it’s important to address those changes now, please let me know, and I’ll make the updates accordingly. 
Otherwise, if everything looks good, could you please approve and merge it?

core/network/src/main/java/com/mifos/core/network/services/ChargeService.kt
niyajali: @itsPronay No it does, create an adapter for that, like this, create the below two files and place them in the utils folder

```
// FlowCallAdapterFactory.kt
import kotlinx.coroutines.flow.Flow
import retrofit2.CallAdapter
import retrofit2.Response
import retrofit2.Retrofit
import java.lang.reflect.ParameterizedType
import java.lang.reflect.Type

class FlowCallAdapterFactory private constructor() : CallAdapter.Factory() {
    override fun get(
        returnType: Type,
        annotations: Array<Annotation>,
        retrofit: Retrofit
    ): CallAdapter<*, *>? {
        if (getRawType(returnType) != Flow::class.java) {
            return null
        }
        check(returnType is ParameterizedType) { "Flow return type must be parameterized as Flow<Foo> or Flow<out Foo>" }
        val responseType = getParameterUpperBound(0, returnType)
        val rawFlowType = getRawType(responseType)
        return if (rawFlowType == Response::class.java) {
            check(responseType is ParameterizedType) { "Response must be parameterized as Response<Foo> or Response<out Foo>" }
            ResponseCallAdapter<Any>(
                getParameterUpperBound(
                    0,
                    responseType
                )
            )
        } else {
            BodyCallAdapter<Any>(responseType)
        }
    }

    companion object {
        @JvmStatic
        fun create() = FlowCallAdapterFactory()
    }
}
```

```
// FlowCallAdapter.kt
import kotlinx.coroutines.flow.Flow
import kotlinx.coroutines.flow.flow
import kotlinx.coroutines.suspendCancellableCoroutine
import retrofit2.Call
import retrofit2.CallAdapter
import retrofit2.Callback
import retrofit2.Response
import java.lang.reflect.Type
import kotlin.coroutines.resume
import kotlin.coroutines.resumeWithException

class ResponseCallAdapter<T>(
    private val responseType: Type
) : CallAdapter<T, Flow<Response<T>>> {
    override fun adapt(call: Call<T>): Flow<Response<T>> {
        return flow {
            emit(
                suspendCancellableCoroutine { continuation ->
                    call.enqueue(object : Callback<T> {
                        override fun onFailure(call: Call<T>, t: Throwable) {
                            continuation.resumeWithException(t)
                        }

                        override fun onResponse(call: Call<T>, response: Response<T>) {
                            continuation.resume(response)
                        }
                    })
                    continuation.invokeOnCancellation { call.cancel() }
                }
            )
        }
    }

    override fun responseType() = responseType
}


class BodyCallAdapter<T>(private val responseType: Type) : CallAdapter<T, Flow<T>> {
    override fun adapt(call: Call<T>): Flow<T> {
        return flow {
            emit(
                suspendCancellableCoroutine { continuation ->
                    call.enqueue(object : Callback<T> {
                        override fun onFailure(call: Call<T>, t: Throwable) {
                            continuation.resumeWithException(t)
                        }

                        override fun onResponse(call: Call<T>, response: Response<T>) {
                            try {
                                continuation.resume(response.body()!!)
                            } catch (e: Exception) {
                                continuation.resumeWithException(e)
                            }
                        }
                    })
                    continuation.invokeOnCancellation { call.cancel() }
                }
            )
        }
    }

    override fun responseType() = responseType
}
```

```kotlin
And use like this
     val retrofit = Retrofit.Builder()
    .baseUrl("https://example.com/")
    .addCallAdapterFactory(FlowCallAdapterFactory.create()) // Add this adapter on your retrofit config
    .build()
```

core/database/src/main/java/com/mifos/room/di/DaoModule.kt
itsPronay: @HekmatullahAmin thanks for the suggestion brother.
but we already used `@InstallIn(SingletonComponent::class)` on the DaoModule, it already ensures that all the provided components in this module are scoped as singletons

HekmatullahAmin: Hi @itsPronay !

Just to clarify the difference between @InstallIn(SingletonComponent::class) and @Singleton in Hilt:

@InstallIn(SingletonComponent::class) means that the dependency will live as long as the application is alive, but it doesn't guarantee a single instance by itself.
@Singleton tells Hilt to reuse the same instance of the dependency within the same scope.

Example:
```
@Module
@InstallIn(SingletonComponent::class)
object NetworkModule {

    @Provides
    fun provideApiService(): ApiService {
        return ApiServiceImpl()
    }
}
```
In this case, every class that requests ApiService gets a new instance, because @Singleton is missing.

Singleton Example:
```
@Module
@InstallIn(SingletonComponent::class)
object NetworkModule {

@Singleton
@Provides
    fun provideApiService(): ApiService {
        return ApiServiceImpl()
    }
}
```

By adding @Singleton, only one instance of SomeService is created and shared across the entire app (within the SingletonComponent scope).
Without @Singleton, multiple instances are created even if SingletonComponent is used.

If you’d like more details, you can check out the official documentation here:
[Hilt - Android Developer Guide](https://developer.android.com/training/dependency-injection/hilt-android)

itsPronay: @HekmatullahAmin Thanks for explaining.  
I have a doubt, since MifosDatabase is already a singleton, wouldn't the DAO instances be shared across the app by default? meaning singleton

therajanmaurya (approved):

PR #2321, opened 2025-02-28
resolved recently created clients fail to open issue

ufuomaisaac:
Fixes - [Jira-#Issue_Number](https://mifosforge.jira.com/browse/MIFOSAC-307)

recently creates clients fail to open issue was resolved

### DIFF
--- a/feature/client/src/main/AndroidManifest.xml
+++ b/feature/client/src/main/AndroidManifest.xml
@@ -15,7 +15,7 @@
     <application>
         <provider
             android:name="androidx.core.content.FileProvider"
-            android:authorities="${applicationId}.provider"
+            android:authorities="com.mifos.mifosxdroid.provider"
             android:exported="false"
             android:grantUriPermissions="true">
             <meta-data
--- a/feature/client/src/main/java/com/mifos/feature/client/clientDetails/ui/ClientDetailsScreen.kt
+++ b/feature/client/src/main/java/com/mifos/feature/client/clientDetails/ui/ClientDetailsScreen.kt
@@ -90,7 +90,6 @@ import androidx.compose.ui.window.DialogProperties
 import androidx.core.content.FileProvider
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
-import androidx.test.core.app.ApplicationProvider.getApplicationContext
 import coil.compose.AsyncImage
 import com.google.accompanist.permissions.ExperimentalPermissionsApi
 import com.google.accompanist.permissions.isGranted
--- a/feature/client/src/main/java/com/mifos/feature/client/clientDetails/ui/ClientDetailsScreen.kt
+++ b/feature/client/src/main/java/com/mifos/feature/client/clientDetails/ui/ClientDetailsScreen.kt
@@ -90,6 +90,7 @@ import androidx.compose.ui.window.DialogProperties
 import androidx.core.content.FileProvider
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.test.core.app.ApplicationProvider.getApplicationContext
 import coil.compose.AsyncImage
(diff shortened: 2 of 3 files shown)


ufuomaisaac:
The is the PR for "recently created clients fail to open" issue

feature/client/src/main/AndroidManifest.xml
```diff
@@ -15,7 +15,7 @@
     <application>
         <provider
             android:name="androidx.core.content.FileProvider"
-            android:authorities="${applicationId}.provider"
+            android:authorities="com.mifos.mifosxdroid.provider"
```
niyajali, line 18: We have configured multiple flavors in our app, such as demo and prod. If we define the package name statically, it will cause platform clashes, making it impossible to install both applications simultaneously. 

And I don't think by changing this it does resolve the issue as indicated in the PR title.

niyajali (changes requested):

niyajali:
@ufuomaisaac Hi, thank you for your efforts. We have completely moved our issue tracking to Jira. Please consider creating a ticket there, similar to how it's done on GitHub, and proceed with your work. You can find the link in the project's README. And join our daily standup for more development update if you haven't joined yet.

niyajali:
There are no issues with the **application configuration files, such as the manifest, ProGuard, R8, or project settings**. All issues are related to the **feature or core** modules, so please focus on working with or modifying those files. Avoid making changes to the configuration files, as these are production apps. If any changes to these files inadvertently violate Play Store policies, we would need to fix them again, which could disrupt the release process.

feature/client/src/main/AndroidManifest.xml
ufuomaisaac: That actually resolved the issue as there was no reference to application id upon clicking it
The error that was resolved was "Couldn't find meta-data for provider with authority com.mifos.mifosxdroid.provider"

If we want to make it dynamic, then application.id has to be declared somewhere in the project.

Could you please advise where the application ID should be defined to make it dynamic?

ufuomaisaac: That actually resolved the issue as there was no reference to application id upon clicking it
The error that was resolved was "Couldn't find meta-data for provider with authority com.mifos.mifosxdroid.provider"

Alright, I will try and focus on resolving the issue within the feature module without making changes to the manifest.
Thank you for your response

niyajali: It sounds like we need to update the package name in the specific function or module where we're handling file-saving functionality. The issue might arise because the package name or cache directory we've defined doesn't match the current flavor or version of the installed app. To resolve this, we should ensure the package name and cache directory to aligns with the app's current configuration.

maybe the issue is in these files not in manifest - https://github.com/openMF/android-client/blob/1bc8f49747d48550e515d30e75c373c93e5dfa49/feature/client/src/main/java/com/mifos/feature/client/createNewClient/CreateNewClientScreen.kt#L300

or here - https://github.com/openMF/android-client/blob/1bc8f49747d48550e515d30e75c373c93e5dfa49/feature/client/src/main/java/com/mifos/feature/client/createNewClient/CreateNewClientScreen.kt#L932

```
private fun Context.createTempImageFile(): File {
    val timeStamp = SimpleDateFormat("yyyyMMdd_HHmmss", Locale.getDefault()).format(Date())
    val imageFileName = "JPEG_${timeStamp}_"
    val storageDir = externalCacheDir ?: cacheDir
    return File.createTempFile(
        imageFileName,
        ".png",
        storageDir
    )
}
```

or updating file_provider or checking the path name might solve this issue here - https://github.com/openMF/android-client/blob/1bc8f49747d48550e515d30e75c373c93e5dfa49/feature/client/src/main/res/xml/feature_client_path_provider.xml

update with existing
```
    <external-cache-path name="external_cache" path="." />
    <cache-path name="internal_cache" path="." />
```

ufuomaisaac: Alright, I will look into that.
Thank you

Issue #410, opened 2016-11-22
Add ability to create group savings account [MIFOSAC-130]


edcable:
**Summary:** 

From the group view, from the more menu, there should be a link to add/create savings account that belongs to a group. The savings account application should contain the same fields as that of a client savings application.

tarun0:
Fixed in #454

PR #2350, opened 2025-03-31
feat(core:domain): Migrate to KMP

biplab1:
Fixes - [Jira-#410](https://mifosforge.jira.com/browse/MIFOSAC-410)


core/domain/build.gradle.kts
```diff
@@ -8,37 +8,67 @@
  * See https://github.com/openMF/android-client/blob/master/LICENSE.md
  */
 plugins {
-    alias(libs.plugins.mifos.android.library)
+    alias(libs.plugins.mifos.kmp.library)
     alias(libs.plugins.mifos.android.library.jacoco)
     alias(libs.plugins.mifos.android.koin)
```
itsPronay: this is for android. Remove this and include `    alias(libs.plugins.mifos.kmp.koin)`

core/domain/build.gradle.kts
```diff
@@ -8,37 +8,67 @@
...
-//    alias(libs.plugins.mifos.android.hilt)
+    alias(libs.plugins.jetbrainsCompose)
+    alias(libs.plugins.compose.compiler)
```
itsPronay, line 15: why do we need this?

itsPronay, line 14: why do we need this!?!

itsPronay (changes requested):
@biplab1 please resolve these

core/domain/build.gradle.kts
biplab1: Implemented the suggestion. Thanks

core/domain/build.gradle.kts
biplab1: Those are added for the following code in build.gradle of core:domain:
```gradle
compose.resources {
    publicResClass = true
    generateResClass = always
    packageOfResClass = "core.domain.generated.resources"
}
```

core/domain/src/commonMain/kotlin/com/mifos/core/domain/useCases/ActivateGroupUseCase.kt
```diff
@@ -23,13 +23,13 @@ class ActivateGroupUseCase(
...
-            trySend(Resource.Loading())
+            emit(Resource.Loading())
             val response = activateRepository.activateGroup(groupId, groupPayload)
-            trySend(Resource.Success(response))
+            emit(Resource.Success(response))
         } catch (exception: Exception) {
-            send(Resource.Error(exception.message.toString()))
+            emit(Resource.Error(exception.message.toString()))
         }
     }
```
itsPronay: @biplab1 don't use `try-catch` inside flow. Take reference from LoginUseCase

biplab1: I will look into it. Thanks.

biplab1: I have implemented your suggestion.

core/domain/src/commonMain/kotlin/com/mifos/core/domain/useCases/GetCenterDetailsUseCase.kt
```diff
@@ -25,20 +25,18 @@ class GetCenterDetailsUseCase(
         centerId: Int,
         genericResultSet: Boolean,
     ): Flow<Resource<Pair<CenterWithAssociations, List<CenterInfo>>>> = flow {
-        try {
-            emit(Resource.Loading())
+        emit(Resource.Loading())
+    }.flatMapLatest {
             repository.getCentersGroupAndMeeting(centerId)
```
itsPronay: when return type is flow, i dont think we need useCases, use it directly in the viewmodel

biplab1: We are returning flow in every useCases. I don't think I understand your statement. Could you please elaborate?

itsPronay: @biplab1 since `repository.getCentersGroupAndMeeting(centerId)`  already returns flow, we do not need to convert it to flow again.

biplab1: I understand your point now. Thanks

core/domain/src/commonMain/kotlin/com/mifos/core/domain/useCases/LoadSavingsAccountsAndTemplateUseCase.kt
```diff
@@ -0,0 +1,45 @@
...
+                val savingProductsAndTemplate = coroutineScope {
+                    val savingsAccount = async { repository.savingsAccounts() }
+                    val template = async { repository.savingsAccountTemplate() }
+
+                    SavingsProductsAndTemplate(
+                        savingsAccount = savingsAccount.await(),
+                        template = template.await(),
+                    )
+                }
+                emit(Resource.Success(savingProductsAndTemplate))
```
itsPronay: use combine instead

biplab1: I will look into it. Thanks

biplab1: I have implemented your suggestion. Thanks again

core/domain/src/commonMain/kotlin/com/mifos/core/domain/useCases/GetCenterDetailsUseCase.kt
biplab1: Here we are combining two flows into a single flow using zip

niyajali (changes requested):
As I can see these most of the  UseCase classes are useless since it doesn't do any Job rather than calling the repository method itself, which we can directly so by accessing the repository in ViewModel. for e.g take a look into this class.
```
class ActivateSavingsUseCase(
    private val repository: SavingsAccountActivateRepository
) {
    operator fun invoke(
        savingsAccountId: Int, 
        request: HashMap<String, String>
    ): Flow<DataState<GenericResponse>> = 
        repository.activateSavings(savingsAccountId, request)
}
```
does it do any extra work rather than calling the method no so it's just useless and anti-pattern and can delete this class and remove overhead.

so my suggestion is to Keep Use Case classes that perform meaningful transformations, business logic, or complex operations. Remove Use Case classes that merely act as pass-through wrappers around repository methods without adding any additional value or processing logic.

Retain Use Cases that:
- Implement complex business rules
- Perform data filtering, sorting, or transformation
- Combine data from multiple repositories
- Add validation or preprocessing steps
- Encapsulate intricate domain-specific logic

Remove Use Cases that:
- Directly delegate to repository methods without any additional processing
- Offer no extra functionality beyond repository method invocation
- Create unnecessary abstraction layers without adding value

biplab1:
@niyajali Thanks for your detailed review. We have a separate ticket for getting rid of useCases which are doing nothing but forwarding a flow. Here is the link to the Jira ticket: https://mifosforge.jira.com/browse/MIFOSAC-440

niyajali (approved):

PR #2352, opened 2025-04-01
feat (core:dataStore) Migrated to KMP

revanthkumarJ:
Fixes - [Jira-#MIFOSAC-330](https://mifosforge.jira.com/browse/MIFOSAC-330)


revanthkumarJ:
@itsPronay I have resolved all errors in this , build is fine but app is crashing saying the Instance is not created for koin. Can you look into it after your common PR

revanthkumarJ:
https://github.com/user-attachments/assets/9a8be7a2-0c35-41bd-948c-10240c6f4634

core/network/src/main/java/com/mifos/core/network/datamanager/DataManagerGroups.kt
```diff
@@ -203,14 +217,16 @@ class DataManagerGroups(
...
+        return when (prefManager.userInfo.firstOrNull()?.userStatus) {
             false -> mBaseApiManager.groupApi.createGroup(groupPayload)
 
             true ->
                 /**
                  * Save GroupPayload in Database table.
                  */
                 databaseHelperGroups.saveGroupPayload(groupPayload)
+
+            null -> TODO()
```
itsPronay: should not be null.

revanthkumarJ: you meant "null -> TODO()" should not be used right?

revanthkumarJ: instead of firstOrNull I can use first() , then there is no issue?

itsPronay: userStatus should not be nullable.. 
plus If its null, the app will crash since you used todo()

itsPronay: > instead of firstOrNull I can use first() , then there is no issue?

What did we use in Mobile-wallet ? you can take reference from there tho first() sounds gd

revanthkumarJ: yah i have changed them to first()

core/model/src/commonMain/kotlin/com/mifos/core/model/objects/users/User.kt
```diff
@@ -10,35 +10,17 @@
...
-            ", base64EncodedAuthenticationKey='" + base64EncodedAuthenticationKey + '\'' +
-            ", authenticated=" + isAuthenticated +
-            ", officeId=" + officeId +
-            ", officeName='" + officeName + '\'' +
-            ", roles=" + roles +
-            ", permissions=" + permissions +
-            '}'
-    }
-
-    companion object {
```
HekmatullahAmin, line 41: You didn't add `AUTHENTICATION_KEY` in your migration, have you checked that this is not needed in the app?

core/network/src/main/java/com/mifos/core/network/BaseApiManager.kt
```diff
@@ -175,8 +175,9 @@ class BaseApiManager(private val prefManager: PrefManager) {
                 prettyPrint = true
             }
 
+            val instanceUrl = prefManager.getServerConfig.value?.getInstanceUrl() ?: "https://dev.mifos.io/fineract-provider/api/v1/"
```
HekmatullahAmin: I think you made `ServerConfig` nullable in UserPreferencesRepository. If nullability was previously handled in the PrefManager. Have a look at it and do the same UserPreferencesRepository. 
Cuause then we will need to this **safe call operator** `?.` everytime

core/network/src/main/java/com/mifos/core/network/MifosInterceptor.kt
```diff
@@ -9,21 +9,22 @@
...
 
     @Throws(IOException::class)
     override fun intercept(chain: Interceptor.Chain): Response {
-        val chianrequest = chain.request()
-        val builder = chianrequest.newBuilder()
-            .header(HEADER_TENANT, prefManager.getServerConfig.tenant)
-        if (prefManager.isAuthenticated()) builder.header(HEADER_AUTH, prefManager.getToken())
+        val chainrequest = chain.request()
+        val builder = chainrequest.newBuilder()
+        val tenant = prefManager.getServerConfig.value?.tenant ?: ""
```
HekmatullahAmin: Same here @revanthkumarJ

core/network/src/main/java/com/mifos/core/network/datamanager/DataManagerCenter.kt
```diff
@@ -120,13 +121,16 @@ class DataManagerCenter(
...
 
     suspend fun createCenter(centerPayload: CenterPayloadEntity?) {
-        when (prefManager.userStatus) {
-            false -> mBaseApiManager.centerApi.createCenter(centerPayload)
-            true ->
-                /**
-                 * Save CenterPayload in Database table.
-                 */
-                centerDatabaseHelper.saveCenterPayload(centerPayload)
+        prefManager.userInfo.collect {
```
HekmatullahAmin: - If we just need the latest value of userInfo, not a continuous stream of updates then we should use `first()` or `firstOrNull()`.
- I assume `userInfo` always emits at least one value. If it might be empty, use `firstOrNull()` and handle the null case.
- Based on code, it looks like `createCenter` is meant to be a one-time operation that checks the current `userStatus` and acts accordingly. Therefore, use instead of collect first is good option.

core/network/src/main/java/com/mifos/core/network/datamanager/DataManagerCenter.kt
```diff
@@ -137,13 +141,15 @@ class DataManagerCenter(
...
     fun getCenterWithAssociations(centerId: Int): Flow<CenterWithAssociations> {
         return flow {
-            when (prefManager.userStatus) {
-                false -> mBaseApiManager.centerApi.getAllGroupsForCenter(centerId)
-                true ->
-                    /**
-                     * Return Groups from DatabaseHelperGroups.
-                     */
-                    centerDatabaseHelper.getCenterAssociateGroups(centerId)
+            prefManager.userInfo.collect { userData ->
```
HekmatullahAmin: @revanthkumarJ brother some code like this doesn't make sense. So it is your job to look where is this function called, how it is called and change accordingly. 
- For example this doesn't emit anything to outer flow.
- Does `mBaseApiManager.centerApi.getAllGroupsForCenter(centerId)` and `centerDatabaseHelper.getCenterAssociateGroups(centerId)` return the same `CenterWithAssociations`?

- If you only need to fetch the data once based on the current userStatus, use userInfo.first() and emit the result:

```
fun getCenterWithAssociations(centerId: Int): Flow<CenterWithAssociations> {
    return flow {
        val userData = prefManager.userInfo.first()
        val result = when (userData.userStatus) {
            false -> mBaseApiManager.centerApi.getAllGroupsForCenter(centerId)
            true -> centerDatabaseHelper.getCenterAssociateGroups(centerId)
        }
        emit(result) // Assuming result is CenterWithAssociations or needs mapping
    }.flowOn(ioDispatcher) // Optional: Run on IO dispatcher
}
```

- If you want the Flow to emit a new CenterWithAssociations every time userStatus changes, use map or flatMapLatest:

```
fun getCenterWithAssociations(centerId: Int): Flow<CenterWithAssociations> {
    return prefManager.userInfo.map { userData ->
        when (userData.userStatus) {
            false -> mBaseApiManager.centerApi.getAllGroupsForCenter(centerId)
            true -> centerDatabaseHelper.getCenterAssociateGroups(centerId)
        }
    }.flowOn(ioDispatcher)
}
```

- I assume the API/database calls are suspend functions returning CenterWithAssociations. If they return a Flow, use flatMapLatest instead

core/network/src/main/java/com/mifos/core/network/di/NetworkModule.kt
```diff
@@ -12,31 +12,41 @@ package com.mifos.core.network.di
...
 import okhttp3.Call
 import okhttp3.OkHttpClient
 import org.koin.android.ext.koin.androidContext
 import org.koin.dsl.module
 import org.mifos.core.apimanager.BaseApiManager
 
 val NetworkModule = module {
-    single { PrefManager(androidContext()) }
+
+    single<UserPreferencesRepository> { UserPreferencesRepositoryImpl(get(), Dispatchers.IO, Dispatchers.Unconfined) }
```
HekmatullahAmin: Don't you have access to common modules Dispatchers use them

core/network/src/main/java/com/mifos/core/network/di/NetworkModule.kt
```diff
@@ -12,31 +12,41 @@ package com.mifos.core.network.di
...
     single { com.mifos.core.network.BaseApiManager(get()) }
 
     single { BaseApiManager }
 
     single {
-        val prefManager: PrefManager = get()
-        val usernamePassword: Pair<String, String> = prefManager.usernamePassword
+        val prefManager: UserPreferencesRepository = get()
+
+        val user = runBlocking { prefManager.userData.first() }
```
HekmatullahAmin: Calling `runBlocking` in a Koin module is problematic. It can block the thread (often the main thread) can lead to performance issues and may cause sth like `Application Not Responding`. 
- use CoroutineScope: 
```
single {
        val prefManager: UserPreferencesRepository = get()
        val baseManager = BaseApiManager.getInstance()
        // Use a scope to configure asynchronously
        CoroutineScope(Dispatchers.IO).launch {
            val user = prefManager.userData.first()
            val serverConfig = prefManager.getServerConfig.first()
            if (serverConfig != null) {
                baseManager.createService(
                    user.username ?: "",
                    user.password ?: "",
                    serverConfig.getInstanceUrl().dropLast(3) ?: "",
                    serverConfig.tenant ?: "",
                    false,
                )
            }
        }
        baseManager
    }
```

feature/auth/src/main/java/com/mifos/feature/auth/login/LoginViewModel.kt
```diff
@@ -103,27 +102,34 @@ class LoginViewModel(
...
     ) {
         // Updating Services
-        baseApiManager.createService(
-            username = username,
-            password = password,
-            baseUrl = prefManager.getServerConfig.getInstanceUrl().dropLast(3),
-            tenant = prefManager.getServerConfig.tenant,
-            secured = false,
-        )
+//        baseApiManager.createService(
```
HekmatullahAmin: I see you commented on some codes in some parts don't you think these comments will cause different functionality

feature/client/src/main/java/com/mifos/feature/client/clientList/presentation/ClientListViewModel.kt
```diff
@@ -46,7 +47,12 @@ class ClientListViewModel(
     }
 
     fun getClientList() {
```
HekmatullahAmin, line 49: - You’re using a coroutine to fetch userStatus, but the subsequent logic doesn’t wait for it. This defeats the purpose of using an asynchronous operation.
- Using a var to store userStatus is unnecessary and error-prone here. Since you only need the value once to make a decision, there’s no need for a mutable variable outside the coroutine.
- Place the conditional logic inside the viewModelScope.launch block to ensure it runs after userStatus is fetched:
```
fun getClientList() {
    viewModelScope.launch {
        val userStatus = prefManager.userInfo.first().userStatus
        if (userStatus) {
            loadClientsFromDb()
        } else {
            loadClientsFromApi()
        }
    }
}
```

feature/client/src/main/java/com/mifos/feature/client/clientSurveySubmit/SurveySubmitViewModel.kt
```diff
@@ -38,7 +40,14 @@ class SurveySubmitViewModel(
     val surveySubmitUiState: StateFlow<SurveySubmitUiState>
         get() = _surveySubmitUiState
 
-    val userId = MutableStateFlow(prefManager.getUserId())
+    var userId = MutableStateFlow(0)
+    init {
```
HekmatullahAmin: Below is better way of doing it if you need reactivity:
```
val userId: StateFlow<Int> = prefManager.userData
        .map { it.userId.toInt() }
        .stateIn(
            scope = viewModelScope,
            started = SharingStarted.WhileSubscribed(5000), // Stop after 5s of no subscribers
            initialValue = 0
        )
```

feature/client/src/main/java/com/mifos/feature/client/syncClientDialog/SyncClientsDialogViewModel.kt
```diff
@@ -65,7 +67,11 @@ class SyncClientsDialogViewModel(
     }
 
     fun syncClient() {
-        if (prefManager.userStatus == Constants.USER_ONLINE) {
+        var userStatus: Boolean = false
```
HekmatullahAmin: @revanthkumarJ Same goes here 

```
fun syncClient() {
    viewModelScope.launch {
        val userStatus = prefManager.userInfo.first().userStatus
        if (userStatus == Constants.USER_ONLINE) {
            checkNetworkConnection {
                syncClientAndUpdateUI()
            }
        }
    }
}
```

- If you have done the same like this in other viewmodels change it, I already mentioned the reason in another comment section

feature/data-table/src/main/java/com/mifos/feature/dataTable/dataTableList/DataTableListViewModel.kt
```diff
@@ -39,10 +40,23 @@ import rx.schedulers.Schedulers
...
+    private val prefManager: UserPreferencesRepository,
     savedStateHandle: SavedStateHandle,
 ) : ViewModel() {
 
+    private var currentUserStatus: Boolean = false
+
+    private val _userStatus = MutableStateFlow(currentUserStatus)
+    val userStatus = _userStatus.asStateFlow()
+
+    init {
```
HekmatullahAmin: Don't do it this way in another comment I mentioned how to do it follow that approach. and change it if you did it this way in other viewmodels

feature/path-tracking/src/main/java/com/mifos/feature/pathTracking/PathTrackingViewModel.kt
```diff
@@ -32,41 +33,52 @@ class PathTrackingViewModel(
...
+        }
+    }
+
     fun refreshCenterList() {
         _isRefreshing.value = true
         loadPathTracking()
         _isRefreshing.value = false
     }
 
     fun loadPathTracking() = viewModelScope.launch(Dispatchers.IO) {
```
HekmatullahAmin, line 54: - Use mifos dispatchers if you have access to common module
- make use DI: 
```
class MyViewModel(
    private val prefManager: UserPreferencesRepository,
    private val getUserPathTrackingUseCase: GetUserPathTrackingUseCase,
    private val ioDispatcher: CoroutineDispatcher
) : ViewModel() {
```

feature/path-tracking/src/main/java/com/mifos/feature/pathTracking/PathTrackingViewModel.kt
```diff
@@ -32,41 +33,52 @@ class PathTrackingViewModel(
...
-        getUserPathTrackingUseCase(prefManager.getUserId()).collect { result ->
-            when (result) {
-                is Resource.Error ->
-                    _pathTrackingUiState.value =
-                        PathTrackingUiState.Error(R.string.feature_path_tracking_failed_to_load_path_tracking)
+        var userId = prefManager.userData.firstOrNull()?.userId
```
HekmatullahAmin: - use val instead of var
- If userId is null, the function silently does nothing. This might be intentional, but it could leave the UI in an uninformative state.
```
fun loadPathTracking() = viewModelScope.launch(Dispatchers.IO) {
    val userId = prefManager.userData.firstOrNull()?.userId
    if (userId != null) {
        getUserPathTrackingUseCase(userId.toInt()).collect { result ->
            when (result) {
                is Resource.Error ->
                    _pathTrackingUiState.value =
                        PathTrackingUiState.Error(R.string.feature_path_tracking_failed_to_load_path_tracking)
                is Resource.Loading -> _pathTrackingUiState.value = PathTrackingUiState.Loading
                is Resource.Success ->
                    result.data?.let { pathTracking ->
                        _pathTrackingUiState.value =
                            if (pathTracking.isEmpty()) {
                                PathTrackingUiState.Error(R.string.feature_path_tracking_no_path_tracking_found)
                            } else {
                                PathTrackingUiState.PathTracking(pathTracking)
                            }
                    }
            }
        }
    } else {
        _pathTrackingUiState.value =
            PathTrackingUiState.Error(R.string.feature_path_tracking_no_user_data)
    }
}
```

feature/path-tracking/src/main/java/com/mifos/feature/pathTracking/PathTrackingViewModel.kt
```diff
@@ -32,41 +33,52 @@ class PathTrackingViewModel(
...
-                                PathTrackingUiState.PathTracking(
-                                    pathTracking,
-                                )
-                            }
-                    }
+                    is Resource.Success ->
+                        result.data?.let { pathTracking ->
+                            _pathTrackingUiState.value =
+                                if (pathTracking.isEmpty()) {
+                                    PathTrackingUiState.Error(R.string.feature_path_tracking_no_path_tracking_found)
```
HekmatullahAmin, line 69: - why we are treating an empty pathTracking list as an Error state?
- Consider making an empty list a valid success state unless the domain requires otherwise
```
is Resource.Success ->
                    _pathTrackingUiState.value = PathTrackingUiState.PathTracking(
                        pathTracking = result.data ?: emptyList()
                    )
```

niyajali:
@revanthkumarJ Is it ready for merge?

feature/path-tracking/src/main/java/com/mifos/feature/pathTracking/PathTrackingViewModel.kt
revanthkumarJ: It is not added by me in this PR, i think changing all feature module files is not the scope of this PR

revanthkumarJ: I think it is not scope of this PR as i haven't added new things here

revanthkumarJ:
> @revanthkumarJ Is it ready for merge?

still doing some changes , i will update once all done

feature/groups/src/main/java/com/mifos/feature/groups/createNewGroup/CreateNewGroupScreen.kt
```diff
@@ -93,7 +93,7 @@ internal fun CreateNewGroupScreen(
         invokeGroupCreation = { groupPayload ->
             viewModel.createGroup(groupPayload)
         },
-        onGroupCreated = { onGroupCreated(it, viewModel.getUserStatus()) },
+        onGroupCreated = { onGroupCreated(it, viewModel.userStatus) },
```
itsPronay: @revanthkumarJ userStatus is a StateFlow. How  you are collecting it here?!
Collect it in the top function and pass it here

revanthkumarJ: sorry i forgot to update it , done now

core/datastore/src/commonMain/kotlin/com/mifos/core/datastore/di/PreferenceModule.kt
```diff
@@ -0,0 +1,37 @@
...
+        )
+    }
+
+    single<UserPreferencesRepository> {
+        UserPreferencesRepositoryImpl(
+            preferenceManager = get(),
+            ioDispatcher = get(named(MifosDispatchers.IO.name)),
+            unconfinedDispatcher = get(named(MifosDispatchers.Unconfined.name)),
+        )
+    }
```
itsPronay: Since it's a repo and we are providing implementation here, use bind instead. See repositoryModule for reference.

core/datastore/src/commonMain/kotlin/com/mifos/core/datastore/model/AppSettings.kt
```diff
@@ -0,0 +1,30 @@
...
+) {
+    companion object {
+        val DEFAULT = AppSettings(
+            tenant = "default",
+            baseUrl = "https://tt.mifos.community/",
+            appTheme = AppTheme.SYSTEM,
+            language = MifosAppLanguage.SYSTEM_LANGUAGE,
+        )
+    }
+}
```
itsPronay, line 30: @niyajali Could you please take a look here? I was wondering when we already have ServerConfig defined, Do we really need to set the baseUrl , tenant again in this instance, or is there a reason it's being redefined here?!

itsPronay (changes requested):

core/network/src/main/java/com/mifos/core/network/datamanager/DataManagerStaff.kt
```diff
@@ -24,29 +25,32 @@ class DataManagerStaff(
...
-                    officeId.toLong(),
-                    null,
-                    null,
-                    null,
-                )
-                    .map(StaffMapper::mapFromEntity)
-            }
+        return prefManager.userInfo.flatMapLatest {
+                userData ->
+            when (userData.userStatus) {
```
itsPronay: could you please format these type of lines? should be this instead
```
        return prefManager.userInfo.flatMapLatest { userData ->
            when (userData.userStatus) {
                false -> flow {

core/network/src/main/java/com/mifos/core/network/di/NetworkModule.kt
itsPronay: @revanthkumarJ implement this

itsPronay:
@revanthkumarJ, please don't resolve conversations on your own. Please let the maintainer resolve them.

core/network/src/main/java/com/mifos/core/network/di/NetworkModule.kt
revanthkumarJ: done

feature/path-tracking/src/main/java/com/mifos/feature/pathTracking/PathTrackingViewModel.kt
revanthkumarJ: Actually there are many places in feature modules where Dispatchers is used  i think that work can be done during that module migration we can do it

core/datastore/src/commonMain/kotlin/com/mifos/core/datastore/di/PreferenceModule.kt
revanthkumarJ: done

core/network/src/main/java/com/mifos/core/network/datamanager/DataManagerStaff.kt
revanthkumarJ: done

core/model/src/commonMain/kotlin/com/mifos/core/model/objects/users/User.kt
revanthkumarJ: it is not used anywhere

core/datastore/src/commonMain/kotlin/com/mifos/core/datastore/model/AppSettings.kt
revanthkumarJ: if we can add passcode, appTheme, language in ServerConfig then we can replace AppSettings

PR #400, opened 2016-11-21
fix: Hide keyboard on opening Navigation Drawer or clicking Search button

tarun0:
Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Apply the `MifosStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #2364, opened 2025-04-22
refactor(core:data): Migrate to KMP

biplab1:
Fixes - [Jira-#400](https://mifosforge.jira.com/browse/MIFOSAC-400)

I would like to thank @itsPronay for the fruitful discussions and his invaluable help in completing this PR.


core/data/src/commonMain/kotlin/com/mifos/core/data/di/RepositoryModule.kt
```diff
@@ -130,13 +130,12 @@ import com.mifos.core.data.repositoryImp.SyncGroupsDialogRepositoryImp
...
 import com.mifos.core.data.repositoryImp.SyncSavingsAccountTransactionRepositoryImp
 import kotlinx.coroutines.CoroutineDispatcher
-import kotlinx.coroutines.Dispatchers
 import org.koin.core.module.dsl.singleOf
 import org.koin.dsl.bind
 import org.koin.dsl.module
 
 val RepositoryModule = module {
-    single<CoroutineDispatcher> { Dispatchers.IO }
+    single<CoroutineDispatcher> { get() }
```
HekmatullahAmin: This will fail at runtime. previously we had discussion i said just look in the Dispatchers module. I had a look we have named CoroutineDispatcher.
This `single<CoroutineDispatcher> { get() }` would work if we had `CoroutineDispatcher` without a name qualifier.

try this:
```
single<CoroutineDispatcher> { get(named(MifosDispatchers.IO.name)) }
```

core/data/src/androidMain/kotlin/com/mifos/core/data/di/AndroidPlatformModule.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    single {
+        AndroidPlatformDependentDataModule(
+            context = androidContext(),
+            dispatcher = get(named(MifosDispatchers.IO.name)),
+        )
+    }
+}
+
+actual val platformModule: Module = AndroidDataModule
```
HekmatullahAmin: Right now we have only one service which is `NetworkMonitor` we can directly bind it, no need to create PlatformDependentDataModule.

- in android we can do sth like below:
```
actual val DataModule: Module = module {
    single<NetworkMonitor> {
        ConnectivityManagerNetworkMonitor(
            context = androidContext(),
            dispatcher = get(named(MifosDispatchers.IO.name)),
        )
    }
}
```
- in other module for example js:
```
class JsNetworkMonitor : NetworkMonitor {
    override val isOnline: Flow<Boolean> = flowOf(true)
}

actual val DataModule: Module = module {
    single<NetworkMonitor> { JsNetworkMonitor() }
}
```

core/network/src/commonMain/kotlin/com/mifos/core/network/services/ClientService.kt
```diff
@@ -61,13 +62,13 @@ interface ClientService {
     ): Flow<Page<ClientEntity>>
 
     @GET(APIEndPoint.CLIENTS + "/{clientId}")
-    suspend fun getClient(@Path("clientId") clientId: Int): ClientEntity
+    fun getClient(@Path("clientId") clientId: Int): Flow<ClientEntity>
```
HekmatullahAmin: .

core/network/src/commonMain/kotlin/com/mifos/core/network/datamanager/DataManagerClient.kt
```diff
@@ -109,7 +110,7 @@ class DataManagerClient(
      * @param clientId for Query in database or REST API request.
      * @return The Client Details
      */
-    suspend fun getClient(clientId: Int): ClientEntity {
+    fun getClient(clientId: Int): Flow<ClientEntity> {
```
HekmatullahAmin: .

core/data/src/commonMain/kotlin/com/mifos/core/data/repository/ClientDetailsRepository.kt
```diff
@@ -9,20 +9,22 @@
...
  */
 interface ClientDetailsRepository {
 
-    suspend fun uploadClientImage(id: Int, file: PartData?)
+    suspend fun uploadClientImage(id: Int, file: PartData)
 
     suspend fun deleteClientImage(clientId: Int)
 
-    suspend fun getClientAccounts(clientId: Int): ClientAccounts
+    fun getClientAccounts(clientId: Int): Flow<DataState<ClientAccounts>>
```
HekmatullahAmin: .

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/ClientDetailsRepositoryImp.kt
```diff
@@ -9,32 +9,42 @@
...
 
-    override suspend fun getClientAccounts(clientId: Int): ClientAccounts {
+    override fun getClientAccounts(clientId: Int): Flow<DataState<ClientAccounts>> {
         return dataManagerClient.getClientAccounts(clientId)
+            .asDataStateFlow()
+            .flowOn(ioDispatcher)
     }
 
-    override suspend fun getClient(clientId: Int): ClientEntity {
+    override fun getClient(clientId: Int): Flow<DataState<ClientEntity>> {
```
HekmatullahAmin: .

HekmatullahAmin:
@biplab1 I see some places you changed suspend to flow which is correct but some places you should keep it suspend.
- if it is one-shot api call like `getClient()` keep it suspend
- if it is observable and list then change the return type to flow
- I just point one example go again and see the same case and change it.

**Problem in Current Code**

Currently, methods like repository.getClient(...)  returning Flow<DataState<T>>, but in GetClientDetailsUseCase, we're treating it as if they were suspend functions:

```
val clientDeferred = async { repository.getClient(clientId) }
```
This leads to a mismatch. Flow emits over time, and you can't await() a Flow directly. While we could use .first() or .collect(), that adds unnecessary overhead when the API call is just a one-shot operation.


Change it like this:
```
@GET("clients/{clientId}")
suspend fun getClient(@Path("clientId") clientId: Int): ClientEntity
```

DataManagerClient:
```
suspend fun getClient(clientId: Int): ClientEntity
```
Repository:
```
suspend fun getClient(clientId: Int): DataState<ClientEntity>
```
RepositoryImpl:
```
override suspend fun getClient(clientId: Int): DataState<ClientEntity> {
    return try {
        DataState.Success(dataManagerClient.getClient(clientId))
    } catch (e: Exception) {
        DataState.Error(e)
    }
}
```
- and I think you are also working on domain once this PR is merged then change usecase accordingly.

biplab1:
@HekmatullahAmin Thanks for your detailed review! I will look into all your suggestions and implement things accordingly.

core/data/src/commonMain/kotlin/com/mifos/core/data/di/RepositoryModule.kt
biplab1: I have implemented your suggestion. Thanks

core/data/src/androidMain/kotlin/com/mifos/core/data/di/AndroidPlatformModule.kt
biplab1: I have implemented your suggestion. Thanks

biplab1:
@niyajali This PR is ready to be merged. Please review and let me know if anything else is required to be done.

biplab1:
@HekmatullahAmin I have addressed all the issues mentioned by you. Please review and let me know if anything else needs to be done.

core/data/src/androidMain/kotlin/com/mifos/core/data/di/AndroidPlatformModule.kt
```diff
@@ -0,0 +1,25 @@
...
+package com.mifos.core.data.di
+
+import com.mifos.core.common.network.MifosDispatchers
+import com.mifos.core.data.util.ConnectivityManagerNetworkMonitor
+import com.mifos.core.data.util.NetworkMonitor
+import org.koin.android.ext.koin.androidContext
+import org.koin.core.qualifier.named
+import org.koin.dsl.module
+
+val AndroidDataModule = module {
```
HekmatullahAmin: @biplab1 Brother there is no need to create another variable you can just assing it directly. I think previously i provided to you.
name it DataModule cause i think it is how for other modules the used for naming convention and then directly assign it to module {}

core/network/src/commonMain/kotlin/com/mifos/core/network/services/CollectionSheetService.kt
```diff
@@ -91,7 +91,7 @@ interface CollectionSheetService {
     ): GenericResponse
 
     @GET(APIEndPoint.CENTERS + "/{centerId}?associations=groupMembers,collectionMeetingCalendar")
-    suspend fun fetchGroupsAssociatedWithCenter(
+    fun fetchGroupsAssociatedWithCenter(
         @Path("centerId") centerId: Int,
-    ): CenterWithAssociations
+    ): Flow<CenterWithAssociations>
```
HekmatullahAmin: This doesn't need to be flow we just retrieve an existing centre. it is called in `FetchGroupsAssociatedWithCenterUseCase` the result of it then be emitted.
image

image

core/network/src/commonMain/kotlin/com/mifos/core/network/services/ClientService.kt
```diff
@@ -61,13 +62,13 @@ interface ClientService {
     ): Flow<Page<ClientEntity>>
 
     @GET(APIEndPoint.CLIENTS + "/{clientId}")
-    suspend fun getClient(@Path("clientId") clientId: Int): ClientEntity
+    fun getClient(@Path("clientId") clientId: Int): ClientEntity
```
HekmatullahAmin: add the suspend back

HekmatullahAmin:
Hi @biplab1 ,

I noticed the recent changes where you've made several API calls return Flow instead of suspend. For the ones that return a list, making them flowable is a good move—thanks for that.

However, for the API calls that were originally suspend and have now been changed to return Flow, I’d suggest reviewing where they’re being called from. In most cases, these are one-time calls used in use cases, and for those, keeping them as suspend is more appropriate and efficient.

For example, `getGroup` in `GroupService.kt` is used in `GroupDetailsViewModel` inside a combine block, which requires a Flow, so making it flowable there makes complete sense.

So, before making an API call return Flow, it would be great if you could first check its usage context. If it's just a one-time value fetch, keeping it as a suspend function would be better.

Thanks a lot

biplab1:
@HekmatullahAmin Thanks for your review and detailed explanation. I will look into your recommendations and update my PR accordingly.

core/network/src/commonMain/kotlin/com/mifos/core/network/services/ClientService.kt
biplab1: Implemented the changes. Thanks

core/network/src/commonMain/kotlin/com/mifos/core/network/services/CollectionSheetService.kt
biplab1: I have implemented the changes. Thanks

core/data/src/androidMain/kotlin/com/mifos/core/data/di/AndroidPlatformModule.kt
biplab1: @HekmatullahAmin I have made the required changes. Thanks

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/CreateNewGroupRepositoryImp.kt
```diff
@@ -9,24 +9,31 @@
...
     private val dataManagerOffices: DataManagerOffices,
     private val dataManagerGroups: DataManagerGroups,
+    private val ioDispatcher: CoroutineDispatcher,
 ) : CreateNewGroupRepository {
 
-    override fun offices(): Flow<List<OfficeEntity>> {
+    override fun offices(): Flow<DataState<List<OfficeEntity>>> {
         return dataManagerOffices.fetchOffices()
+            .asDataStateFlow()
+            .flowOn(ioDispatcher)
```
HekmatullahAmin: @biplab1 good job you PR looks good now to me. But i have one suggestion. 
- I see you added `.flowOn(ioDispatcher)` however it is not causing a problem but it is redundant. Because when we use libraries like ktor, retrofit, room and etc.. they are  somewhere behind run these calls in right thread.
- I suggest you have a look at this youtube video it gives a good explaination: https://www.youtube.com/watch?v=J5h0IJs8aV0

biplab1:
@HekmatullahAmin Here is the summary of the changes that have been made related to Flow:
1. Return a `Flow<List<T>>`, if an API call is returning a `List<T>` (in APIs and Services)
2. Ensure that one-shot API calls do not return a Flow
3. In some places, where a Flow is required for a one-shot API call (for example, a combine block in a ViewModel), the one-shot API call (in services or APIs) has been kept intact (i.e., as a suspend).  Instead, it is returned as a Flow from the corresponding DataManager.
4. In the repository layer of `core/data` module
  a. Return a `Flow<DataState<T>>`, if the function call is returning `Flow<T>`
  b. Return a `Flow<DataState<List<T>>>`, if the function call is returning `List<T>`

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/CreateNewGroupRepositoryImp.kt
biplab1: @HekmatullahAmin I will take a look and update my PR accordingly. Thanks again!

core/data/src/androidMain/kotlin/com/mifos/core/data/util/ConnectivityManagerNetworkMonitor.kt
```diff
@@ -66,7 +66,8 @@ internal class ConnectivityManagerNetworkMonitor(
         awaitClose {
             connectivityManager.unregisterNetworkCallback(callback)
         }
-    }.flowOn(ioDispatcher)
+    }
+        .flowOn(ioDispatcher)
```
HekmatullahAmin, line 70: for example this is good example that you added .flowOn(ioDispatcher).
flowOn(ioDispatcher) pushes all of that upstream logic off the main thread and the whole callbackFlow {} setup runs on Dispatchers.IO

biplab1: I just saw this comment of yours. Thanks for finding this example.

HekmatullahAmin:
@niyajali , @therajanmaurya, this PR looks good to me. If you don't have any suggestions, merge it.

niyajali:
@HekmatullahAmin Mark your conversations as Resolved

niyajali (approved):

PR #445, opened 2016-12-14
Fix: Issue #403 - Dynamically updated nominal interest rate 

tarun0:
Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Apply the `MifosStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #2380, opened 2025-05-19
feat(feature:client): add address fields to create client form

biplab1:
Fixes - [Jira-#445](https://mifosforge.jira.com/browse/MIFOSAC-445)

* With enable-address = true (Global configuration)

https://github.com/user-attachments/assets/af4ec2ec-f98d-44fd-9b11-03d9390c9b01

* With enable-address = false (Global configuration)

https://github.com/user-attachments/assets/0217c6dd-169d-489a-8a22-8fdc9b3710b6

I would like to thank @itsPronay, @HekmatullahAmin and @niyajali for their valuable input and discussions at each stage of progress in resolving the issue.

- [ X] Run the static analysis check `./gradlew check` or `ci-prepush.sh` to make sure you didn't break anything

- [ X] If you have multiple commits please combine them into one commit by squashing them.


niyajali:
@biplab1 Show all address fields only when the user checked the address active checkbox. And rename the field name to "Enable/Provide/Use Address" instead of Address active.

biplab1:
> @biplab1 Show all address fields only when the user checked the address active checkbox. And rename the field name to "Enable/Provide/Use Address" instead of Address active.

@niyajali I looked at the web-app implementation which the android-client follows. It handles the steps for the address field in the create client form in the following way:

1. The address field shows up only when the admin enables address by toggling the enable-address button in Global Configuration. 
2. When the enable-address is enabled (enable-address = true), the address fields are shown and the server won't accept the create client form without the address. 
3. If the enable-address is disabled (enable-address = false), the address form/section do not show in the create client form.

The web-app has a option for adding multiple addresses for a single client, however, as @therajanmaurya pointed out that we can go for a single address in the android-client implementation for now.

Regarding the "Address Active" checkbox name, it actually represents the "Active Status" of the address whether the client currently uses that address. It can be marked true or false from the web-app by the filed officer or the admin. Instead of mentioning just "Active Status" immediately below the address fields, I mentioned it to be "Address Active" to remove any ambiguity with the "Client Active" checkbox which is actually "Active?" in web-app create client form. 

Below, I have attached screenshots of web-app, where it shows how the create client form looks with enable-address (true and false), a address form and the "Active Status" (address field) in the web-app:

1. Create client form for enable-address = false (by admin) in Configuration Wizard -> System Setup -> Configurations -> enable-address -> disable. Address section is absent.

create_client_form_enable_address_false

2. Create client form for enable-address = true (by admin) in Configuration Wizard -> System Setup -> Configurations -> enable-address -> enable. Address section present as can be seen in the top right corner.

create_client_form_enable_address_true

3. An address form is displayed on pressing "+Add":

address_form

4. Address form "Active Status" after adding address fields in the form:

Address_form_Active_Status

biplab1:
Just to clarify, a field officer doesn't have admin rights. He cannot (and shouldn't) change the Global Configurations settings.

niyajali (approved):

PR #2403, opened 2025-06-10
fix: Settings Module Enhancements (Theme Change and Clean Up)

revanthkumarJ:
Fixes - [Jira-#450](https://mifosforge.jira.com/browse/MIFOSAC-450)

https://github.com/user-attachments/assets/82d2c2ca-8d85-4280-b070-2db1f73bb7b0


feature/settings/src/commonMain/kotlin/com/mifos/feature/settings/settings/SettingsScreen.kt
```diff
@@ -152,23 +161,22 @@ internal fun SettingsScreen(
...
+        SyncSurveysDialog(
+            closeDialog = {
+                showSyncSurveyDialog = false
+            },
+        )
     }
 
     if (showServerConfig) {
-        ModalBottomSheet(
+        Dialog(
```
itsPronay, line 172: any reason we replacing bottmSheet with dialog?

revanthkumarJ: specific reason was not there in dev branch app i seen dialog and in bottom sheet it is difficult for typing like if entered something everytime i have to click back button again move bottom drawer up.

If we get new mockups we can do accordingly .

revanthkumarJ:
@itsPronay any other changes?

cmp-navigation/src/commonMain/kotlin/cmp/navigation/ComposeApp.kt
```diff
@@ -9,25 +9,68 @@
...
+            MAIN_GRAPH
+        } else {
+            AUTH_GRAPH
+        }
+    }
+
+    when (uiState) {
+        is MainUiState.Loading -> {
+            MifosCircularProgress(Modifier.fillMaxWidth())
+        }
```
itsPronay: should it not be splashScreen instead? when the mainUi is in loading state

revanthkumarJ: i think yes it should be splash screen i will update

revanthkumarJ: done

biplab1 (approved):

niyajali:
@itsPronay review this one

cmp-navigation/src/commonMain/kotlin/cmp/navigation/ComposeApp.kt
```diff
@@ -9,25 +9,69 @@
...
+            MifosTheme(isDarkTheme) {
+                RootNavGraph(
+                    networkMonitor = networkMonitor,
+                    navHostController = navController,
+                    startDestination = navDestination,
+                    onClickLogout = {
+                        viewModel.logout()
+                        navController.navigate(AUTH_GRAPH) {
+                            popUpTo(navController.graph.id) {
+                                inclusive = true
```
itsPronay: @revanthkumarJ , can you attach a updated video of splash screen ?  as I can see from the logic, the splashScreen never gets called. 

cause when `MainUiState.Success` is true - you are showing content and the splash screen is inside that content.
and at the same time, navDestination value is also changing which doesn't match to splash screen

when `        is MainUiState.Loading -> SplashScreens.SplashScreenRoute.route` The UI is just a loading screen, splash never gets called

feature/settings/src/commonMain/kotlin/com/mifos/feature/settings/settings/SettingsScreen.kt
itsPronay: okay.. just make sure that items fit in that dialog/make it scrollable

cmp-navigation/src/commonMain/kotlin/cmp/navigation/ComposeApp.kt
revanthkumarJ: as the time is so minimal it is directly opening either home screen or auth

revanthkumarJ: yes according to the logic splash is won't called and i am using is MainUiState.Loading -> {
            MifosCircularProgress(Modifier.fillMaxWidth())
        }
        
 no matter what i keep here is MainUiState.Loading -> SplashScreens.SplashScreenRoute.route i won't matter i think

itsPronay: it is not because of time revant, add a delay and see

revanthkumarJ:
https://github.com/user-attachments/assets/24936370-f5bf-4bb7-8369-9d24bdbb9d8b

revanthkumarJ:
@itsPronay now splash screen is showing while state is loading

itsPronay:
@revanthkumarJ we are supposed to replace first splash screen with our own custom splash screen right? but here it is showing 2 splash screen if you notice. 


> @itsPronay now splash screen is showing while state is loading

@revanthkumarJ we are supposed to replace the first splash screen with our own custom splash screen right? but here it is showing 2 splash screens if you notice.

revanthkumarJ:
@itsPronay ok i will remove the first one then

biplab1:
@revanthkumarJ We’re looking to verify the navigation flow and ensure that navigation to all features is working correctly as expected.
Could you please upload a screen recording showing the complete flow—starting from the splash screen and login, and covering all features that have been migrated to CMP with an existing navigation route?

revanthkumarJ:
https://github.com/user-attachments/assets/e877b2e6-0840-4fd0-9e7e-9eb2c0260a73

@biplab1 this is the latest video  
For clarity what achieves in this pr:  
* uncommented the code in settings module which was commented by Aditya due to some issues at that time  
* implemented theming  
* previously while the MainUiState loading login screen comes for 5 to 10 sec now replaced with loader  
* removed the splash module after asking niyaj brother as other projects only use Theme.xml and no custom splash module is needed  
* back icon navigation issue will be solved by Pronay

revanthkumarJ:
@itsPronay can you review once

cmp-navigation/src/commonMain/kotlin/cmp/navigation/ComposeApp.kt
```diff
@@ -9,24 +9,64 @@
...
+    val uiState by viewModel.uiState.collectAsStateWithLifecycle()
+    val navController = rememberNavController()
+
+    val navDestination = when (uiState) {
+        is MainUiState.Loading -> LOADING_GRAPH
+        is MainUiState.Success -> if ((uiState as MainUiState.Success).isAuthenticated) {
+            MAIN_GRAPH
+        } else {
+            AUTH_GRAPH
+        }
```
itsPronay: why do we need `LoadingGraph` when we removed `splashModle`,  instead it should be 
```kt
val navDestination = when (uiState) {
        is MainUiState.Loading -> AUTH_GRAPH
        is MainUiState.Success -> if ((uiState as MainUiState.Success).isAuthenticated) {
            MAIN_GRAPH
        } else {
            AUTH_GRAPH
        }
```

See mobile wallet for reference

cmp-navigation/src/commonMain/kotlin/cmp/navigation/navigation/RootNavGraph.kt
```diff
@@ -9,47 +9,53 @@
...
-        splashNavGraph(
-            navigateLogin = navHostController::navigateToLogin,
-            navigatePasscode = {},
-        )
+        composable(LOADING_GRAPH) {
+            MifosCircularProgress(
+                text = "Loading",
+                modifier = Modifier.fillMaxSize(),
+            )
+        }
```
itsPronay: same here. This part is extra.

itsPronay (changes requested):

cmp-navigation/src/commonMain/kotlin/cmp/navigation/navigation/RootNavGraph.kt
revanthkumarJ: removed

cmp-navigation/src/commonMain/kotlin/cmp/navigation/ComposeApp.kt
revanthkumarJ: done

itsPronay (approved):

niyajali (approved):

biplab1:
@revanthkumarJ Thanks for clarifying. Since this PR is focused on the settings module, the video makes sense for that scope.
We can revisit a full navigation flow check once all relevant features are integrated and ready for review.

PR #2412, opened 2025-06-24
fix: failed to create savings account

itsPronay:
Fixes - https://mifosforge.jira.com/browse/MIFOSAC-471

https://github.com/user-attachments/assets/e8349e25-31fb-4f40-9e50-f690479e724d


revanthkumarJ (approved):

biplab1:
Looks good to me. I noticed two notifications showing recent savings account being created. Screenshot attached for reference.

savings-account-creation-notification

biplab1 (approved):

core/model/src/commonMain/kotlin/com/mifos/core/model/objects/payloads/SavingsPayload.kt
```diff
@@ -9,9 +9,12 @@
...
 package com.mifos.core.model.objects.payloads
 
+import kotlinx.serialization.Serializable
+
 /**
  * Created by nellyk on 2/10/2016.
  */
+@Serializable
 class SavingsPayload {
     var productId: Int? = null
```
Nagarjuna0033, line 19: @itsPronay are we still using var

itsPronay: we havent refactored the model classes yet, all the classes in the model package are still using var. 
only the room entity classes were refactored earlier while doing room migration

niyajali (approved):

Issue #475, opened 2017-01-22
Incorrect content in client screen, if data are not available.


k4wel:
**Summary:** 

3 same rows ''Saving Accounts' in client screen.

**Steps to reproduce:** 

1. Go to the list of clients.
2. Turn off the Internet
3. Select any client

**Expected behavior:** 

Normally, we should see lists of loan, saving and recurring accounts.  And even when data are not available, we should see the same with zeros.

**Observed behavior:** 

We see 3 lines "Saving accounts"

**Screenshots:** 

![screenshot_20170122-135931](https://cloud.githubusercontent.com/assets/22119535/22182326/90ed1f66-e0ab-11e6-8c4b-ad0cdb49927e.png)

mayank-kgp:
@TypicalGitHubUser  Are you working on it or should I try?

TypicalGitHubUser:
@mayank-kgp I think you can try. I`m not sure I can take care of it in the short run.

mayank-kgp:
@TypicalGitHubUser I think we can't see the details because details would be fetched from API when we click on any client from the list.  If we turn off the internet before clicking on any item, the details are not available. 
Please give a comment.

TypicalGitHubUser:
@mayank-kgp 3 lines with exatcly the same name instead of 3 diffferent looks odd and may create negative impression of app. Either correct names or hide them at all would be more appropriate.

mayank-kgp:
I think we should take a suggestion of @therajanmaurya

edcable:
Just confirming that I'm facing the same issue. I will let @therajanmaurya make the final call but agree with @TypicalGitHubUser that it shouldn't display 3 rows of savings accounts but rather the same account headings you would see if online, simply with no data.

therajanmaurya:
@TypicalGitHubUser you are right. If network request failed for some reason. App show should 
"Failed to fetch client details and account." and hide the whole UI and show instead of an ImageView and TextView with a message and on swipe refresh user can make the request again for data.

If anyone wants to pick who can. Please be careful ClientDetailsFragment code.

mayank-kgp:
@therajanmaurya I would like to work on this issue.
Any suggestion or something that you want to tell about before working?

therajanmaurya:
@mayank-kgp Go ahead.

hardik124:
@therajanmaurya  i would like to work on this.

Kigamba:
I would like to work on this @edcable @therajanmaurya

[...]

PR #2413, opened 2025-06-25
fix : Group Module (api), added Navigations in Feature Nav Host

revanthkumarJ:
Fixes - [Jira-#475](https://mifosforge.jira.com/browse/MIFOSAC-475)

What achieved in this PR:
on clicking group clients getting serializable error , added serailizable tag
navigations are missing in Feature NavHost : Added all of them
on clicking settings instance url after updating , app crashing - resolved
Group List Item not same as dev branch so replaced with dev code

image


cmp-navigation/src/commonMain/kotlin/cmp/navigation/navigation/FeatureNavHost.kt
```diff
@@ -113,7 +126,7 @@ internal fun FeatureNavHost(
             addGroupLoanAccount = appState.navController::navigateToGroupLoanScreen,
             addSavingsAccount = appState.navController::navigateToAddSavingsAccount,
             loadDocumentList = appState.navController::navigateToDocumentListScreen,
-            clientListFragment = {},
+            clientListFragment = { _ -> appState.navController.navigateToClientListScreen() },
```
itsPronay: We don't need to pass data if we are not using it

feature/groups/src/commonMain/kotlin/com/mifos/feature/groups/groupList/GroupsListScreen.kt
```diff
@@ -85,22 +89,42 @@ fun GroupItem(
...
+                    { Text(text = it) }
+                },
+                trailingContent = {
+                    Row(
+                        verticalAlignment = Alignment.CenterVertically,
+                        horizontalArrangement = Arrangement.spacedBy(4.dp),
+                    ) {
+                        if (group.sync) {
+                            Icon(imageVector = MifosIcons.DoneAll, contentDescription = "Sync")
+                        }
```
itsPronay: i think using a default value would be better instead of doing 
`
data?.let {
                    { Text(text = it) }
}
`


we can use this instead
                    `{ Text(text = it ?: "null or something else") }`

revanthkumarJ: like if accountNumber is null 

it won't show

what is the need of saying account Number not exists like that?

itsPronay: if vales are null, it will affect `ListItem` card size (when value is null and not null), they would look different, we can use a default value to make the size consistent.

itsPronay (approved):

Nagarjuna0033 (approved):

Aditya3815 (approved):

cmp-navigation/src/commonMain/kotlin/cmp/navigation/navigation/FeatureNavHost.kt
```diff
@@ -68,19 +74,26 @@ internal fun FeatureNavHost(
 
         dataTableNavGraph(
             navController = appState.navController,
-            clientCreated = { _, _ -> },
+            clientCreated = { client, userStatus ->
+                appState.navController.popBackStack()
+                appState.navController.popBackStack()
```
niyajali: Multiple navigation

cmp-navigation/src/commonMain/kotlin/cmp/navigation/navigation/RootNavGraph.kt
```diff
@@ -46,6 +47,7 @@ fun RootNavGraph(
                 modifier = modifier,
                 networkMonitor = networkMonitor,
                 onClickLogout = onClickLogout,
+                navigateToLogin = { navHostController.navigateToLogin() },
```
niyajali: Use method reference(navcontroller::navigate to login)

feature/groups/src/commonMain/kotlin/com/mifos/feature/groups/groupList/GroupsListScreen.kt
```diff
@@ -85,22 +89,44 @@ fun GroupItem(
...
-                    text = it,
-                    style = MaterialTheme.typography.titleSmall,
-                )
+            ListItem(
+                leadingContent = {
+                    Canvas(
+                        modifier = Modifier.size(16.dp),
+                        onDraw = {
+                            drawCircle(
+                                color = if (group.active == true) Color.Green else Color.Red,
```
niyajali: Why the colour has been hardcoded use M3 color

niyajali (changes requested):

cmp-navigation/src/commonMain/kotlin/cmp/navigation/navigation/FeatureNavHost.kt
revanthkumarJ: I am not sure of it, i  copied left over navigations of feature NavHost  from development branch and checking them , currently can't check it because it is depended on create client which is currently still doing by biplab.

so i added a TODO to check it while doing create client ticket

cmp-navigation/src/commonMain/kotlin/cmp/navigation/navigation/RootNavGraph.kt
revanthkumarJ: updated

feature/groups/src/commonMain/kotlin/com/mifos/feature/groups/groupList/GroupsListScreen.kt
revanthkumarJ: updated

cmp-navigation/src/commonMain/kotlin/cmp/navigation/navigation/FeatureNavHost.kt
niyajali: Remove one backstack navigation

revanthkumarJ: @niyajali done

niyajali (approved):

PR #478, opened 2017-01-24
feat : Path Tracking Using FusedLocationApi and showing result in google maps

therajanmaurya:
feat : Path Tracking Using FusedLocationApi and showing result in google maps

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Apply the `MifosStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #2417, opened 2025-06-26
fix(feature:client): fix signature reset issue and enforce white canvas background

biplab1:
Fixes - [Jira-#478](https://mifosforge.jira.com/browse/MIFOSAC-478)

Change Summary:

- Forced canvas background to white in all themes to ensure black ink signatures are visible and export-ready.
- Fixed reset button not clearing the canvas by using a shared DrawingState and adding a clear() method.

Before Fix (Dark Mode):

https://github.com/user-attachments/assets/06779685-8240-4281-8397-45f7d2ee01b0

After Fix (Dark Mode):

https://github.com/user-attachments/assets/849ff681-6aba-4b99-9426-ef1a73ab600c

Light Mode: 

https://github.com/user-attachments/assets/d0bc05e2-43d2-41bd-82e3-b2d670d36ba6

Signature Reset and Upload from Gallery:

https://github.com/user-attachments/assets/2cd62890-25b7-4b2d-8b79-e97b5a4a16b4

- [X] If you have multiple commits please combine them into one commit by squashing them.

### DIFF
--- a/feature/client/src/androidMain/kotlin/com/mifos/feature/client/clientSignature/SignatureScreen.android.kt
+++ b/feature/client/src/androidMain/kotlin/com/mifos/feature/client/clientSignature/SignatureScreen.android.kt
@@ -32,7 +32,6 @@ import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableIntStateOf
-import androidx.compose.runtime.mutableStateListOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
@@ -40,7 +39,6 @@ import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.geometry.Rect
 import androidx.compose.ui.graphics.Color
-import androidx.compose.ui.graphics.Path
 import androidx.compose.ui.graphics.asAndroidBitmap
 import androidx.compose.ui.graphics.asImageBitmap
 import androidx.compose.ui.layout.boundsInRoot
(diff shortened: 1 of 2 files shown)


feature/client/src/androidMain/kotlin/com/mifos/feature/client/clientSignature/SignatureScreen.android.kt
```diff
@@ -77,7 +78,7 @@ internal actual fun SignatureScreen(
     val drawColor = Color.Black
     val drawBrush = 5f
 
-    val paths = remember { mutableStateListOf<PathState>() }
+    val drawingState = remember { DrawingState() }
```
itsPronay, line 81: always use `remembersavable`. `remember` doesn't survive configuration changes

biplab1: Thanks for the suggestion! While `rememberSaveable` works great for primitive types (which it handles automatically), `DrawingState` uses non-serializable types like `Path`, so it would require a custom `Saver`. I explored that, but given users can easily re-sign if the screen rotates, the added complexity doesn’t offer much real value. Keeping it simple for now.

Reference: [UI Logic with rememberSaveable](https://developer.android.com/develop/ui/compose/state-saving#ui-logic)
>rememberSaveable stores [UI element state](https://developer.android.com/topic/architecture/ui-layer/>stateholders#ui-state) in a [Bundle](https://developer.android.com/reference/android/os/Bundle) through the saved instance state mechanism.
>
>It is able to store primitive types to the bundle automatically. If your state is held in a type that is not primitive, like a data class, you can use different storing mechanisms, such as using the [Parcelize](https://developer.android.com/>kotlin/parcelize) annotation, using Compose APIs like [listSaver](https://developer.android.com/reference/kotlin/>androidx/compose/runtime/saveable/package-summary#listSaver(kotlin.Function2,kotlin.Function1)) and [mapSaver](https://developer.android.com/reference/kotlin/androidx/compose/runtime/saveable/package->summary#mapSaver(kotlin.Function2,kotlin.Function1)), or implementing a custom saver class extending Compose runtime [Saver](https://developer.android.com/reference/kotlin/androidx/compose/runtime/saveable/Saver) class.

revanthkumarJ (approved):

itsPronay (approved):

niyajali (approved):
@biplab1 you can try this CMP library - https://github.com/niyajali/compose-signature

biplab1:
@niyajali Thanks for the suggestion! I'll check out the CMP library and see if it fits our use case.

PR #2419, opened 2025-06-26
Fix: Failed to activate centers/groups/clients

itsPronay:
Fixes -https://mifosforge.jira.com/browse/MIFOSAC-479

https://github.com/user-attachments/assets/837d2186-4016-48e7-b509-b351a5b96a70

### DIFF
--- a/core/domain/src/commonMain/kotlin/com/mifos/core/domain/useCases/ActivateGroupUseCase.kt
+++ b/core/domain/src/commonMain/kotlin/com/mifos/core/domain/useCases/ActivateGroupUseCase.kt
@@ -13,29 +13,23 @@ import com.mifos.core.common.utils.DataState
 import com.mifos.core.data.repository.ActivateRepository
 import com.mifos.core.data.util.extractErrorMessage
 import com.mifos.core.model.objects.clients.ActivatePayload
-import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.flow
 
 class ActivateGroupUseCase(
     private val activateRepository: ActivateRepository,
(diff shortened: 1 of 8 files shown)


revanthkumarJ (approved):

core/domain/src/commonMain/kotlin/com/mifos/core/domain/useCases/ActivateGroupUseCase.kt
```diff
@@ -11,17 +11,25 @@ package com.mifos.core.domain.useCases
...
+        val response = activateRepository.activateGroup(groupId, groupPayload)
+
+        if (response.status.value != 200) {
+            val errorMessage = extractErrorMessage(response)
+            DataState.Error(Exception(errorMessage), null)
+        } else {
+            DataState.Success(Unit)
+        }
+    } catch (e: Exception) {
+        DataState.Error(e)
```
biplab1, line 33: You may try using `MFErrorParser.kt`
```kotlin
MFErrorParser.errorMessage(e)
```

biplab1 (approved):

core/domain/src/commonMain/kotlin/com/mifos/core/domain/useCases/ActivateGroupUseCase.kt
itsPronay: @biplab1, if you notice in the viewModel, we are using `stringResource` as error message instead of `Datastate.Error`. We don't need it since we are not using it

niyajali (approved):

PR #2420, opened 2025-06-28
fix : Create Client

revanthkumarJ:
Fixes - [Jira-#469](https://mifosforge.jira.com/browse/MIFOSAC-469)

https://github.com/user-attachments/assets/b060ac38-b231-4e91-85dc-36d5cb2fc308


feature/client/src/commonMain/kotlin/com/mifos/feature/client/createNewClient/CreateNewClientViewModel.kt
```diff
@@ -115,13 +124,16 @@ class CreateNewClientViewModel(
         }
     }
 
-    fun uploadImage(id: Int, imageFile: PlatformFile) {
+    fun uploadImage(id: Int) {
```
biplab1, line 127: Previously, the `uploadImage` function accepted the `PlatformFile` directly, but now the logic is split into `uploadImage(id: Int)` and `updateSelectedImage(image: PlatformFile?)`. Could you clarify the reason for separating image selection from the upload? Just want to understand the benefit of this change.

revanthkumarJ: moved this conversation to slack

biplab1 (approved):
The changes look good to me. This can be merged.

itsPronay (approved):

niyajali (approved):

Issue #483, opened 2017-01-26
each item takes full height of the screen 


dilpreet96:
**Summary:** 

Each list-item for ClientNameListAdapter,GroupNameListAdapter,CentersListAdapter takes the full height of the screen. Each list-item should take enough height of its content.

**Steps to reproduce:** 

Click on Clients or Groups or Centers from the Navigation Drawer. Each list-item takes full length of screen

**Expected behavior:** 

list-item should take enough height of its content .

**Observed behavior:** 

 list-item takes full length of screen

**Device and Android version:** 

Samsung on-7
Android version: 6.0.1

**Screenshots:** 
<img src="http://i.imgur.com/j34aRW3.png" style="width:100px; height:200px;">

therajanmaurya:
@dilpreet96 are you working on latest build?
Can you please test in other device and If you are getting the same issue or not?

dilpreet96:
@therajanmaurya ,Yes I am working on latest build . (Checked again by fetching from upstream/master) 
Got same result on Micromax Canvas 2 Colours ,
Android Version:4.4.2
Shall I work on it ?

therajanmaurya:
It's strange because I am working on the latest build and I can't reproduce this issue. 
My device is Moto G turbo 3 API 24

therajanmaurya:
@edcable @tarun0 @mayank-kgp Can you please confirm about this issue.

mayank-kgp:
@therajanmaurya App is working fine on my device. My device is Micromax yureka with android version 5.1.1

therajanmaurya:
@dilpreet96 Ok If you are really getting this issue, send a PR and let me look closely why this is happening.

dilpreet96:
@therajanmaurya PR sent.
<View/> in row_client_name and row_group_name where causing problem even after making the parent as wrap_content so removed the <View/> and added the isSelected() logic to linearLayout background for multiple selections.

tarun0:
I too observed this behaviour once. However, I could not replicate the issue after one or two times and it was like 1-2 months ago and never observed that again.
Tried again now with Kitkat and Marshmallow and the app works fine for me as well.

therajanmaurya:
fixed in #533 and #534

Issue #480, opened 2017-01-25
Long press on center view brings up calendar


edcable:
**Summary:** 

Do a long press on any center form the list of centers and it brings up the date picker 
**Steps to reproduce:** 

See above

**Expected behavior:** 

Not sure really but typically you could select a group or client for syncing. 

**Observed behavior:** 

Date picker gets displayed 
**Device and Android version:** 

Le Eco S3 with Android 6.0

**Screenshots:** 

![screenshot_20170124-153651](https://cloud.githubusercontent.com/assets/823454/22273227/4430abde-e254-11e6-874c-6afbb282ae41.jpg)

tarun0:
The expected behaviour, most probably, should be to check visits and update payments on that day. Here's some related info: https://mifosforge.jira.com/wiki/display/docs/Working+with+center+lists

therajanmaurya:
@edcable right now center feature does not have support for syncing the center and their group's client. It only shows a dialog picker to pick date and check their is any meeting is schedule or not.

therajanmaurya:
I think this is not an issue. This was the small discussion about the Long press on center view.
Closing this issue for know.

PR #2421, opened 2025-06-28
fix : Client Image (Edit , Delete Dialog)

revanthkumarJ:
Fixes - [Jira-#480](https://mifosforge.jira.com/browse/MIFOSAC-480)
Fixes - [Jira-#483](https://mifosforge.jira.com/browse/MIFOSAC-483)

https://github.com/user-attachments/assets/13f45690-3246-434b-b89b-04610f9d932b


feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientDetails/ClientDetailsViewModel.kt
```diff
@@ -130,23 +125,20 @@ class ClientDetailsViewModel(
         }
     }
 
-    fun saveClientImage(clientId: Int, imageFile: PlatformFile) {
+    fun saveClientImage(clientId: Int, imageFile: PlatformFile?) {
+        if (imageFile == null)return
```
biplab1: Looks like there’s still a missing space before the `return` statement — please fix that.

core/domain/src/commonMain/kotlin/com/mifos/core/domain/useCases/UploadClientImageUseCase.kt
```diff
@@ -26,13 +27,10 @@ class UploadClientImageUseCase(
...
-            emit(DataState.Loading)
+    operator fun invoke(id: Int, image: MultiPartFormDataContent): Flow<DataState<String>> {
+        return flow {
             repository.uploadClientImage(id, image)
-            DataState.Success(getString(Res.string.core_domain_client_image_uploaded_successfully))
-        } catch (e: Exception) {
-            DataState.Error(e)
-        }
+            emit(getString(Res.string.core_domain_client_image_uploaded_successfully))
+        }.asDataStateFlow()
```
HekmatullahAmin: @revanthkumarJ `asDataStateFlow` doesn't work this way. Right now, `repository.uploadClientImage(...)` could fail silently, but Success is always emitted.
our `asDataStateFLow` takes a flowable and return three state: Success, Loading or Error.

so the previous logic was right
```
operator fun invoke(id: Int, image: MultiPartFormDataContent): Flow<DataState<String>> = flow {
    emit(DataState.Loading) // Always emitted first

    try {
        repository.uploadClientImage(id, image)
        emit(DataState.Success(getString(Res.string.core_domain_client_image_uploaded_successfully)))
    } catch (e: Exception) {
        emit(DataState.Error(e))
    }
}
```

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientDetails/ClientDetailsViewModel.kt
```diff
@@ -130,23 +125,20 @@ class ClientDetailsViewModel(
...
         try {
-            val bytes = FileKit.compressImage(
-                file = imageFile,
-                imageFormat = ImageFormat.PNG,
-                quality = 100,
-            )
-            val outFile = FileKit.filesDir / "client_image_$clientId.png"
-            outFile.write(bytes)
+            _showLoading.value = true
+            val outFile = compressImage(imageFile, clientId)
```
HekmatullahAmin: this `saveAutoClientImage` is only used inside the `saveClientImage` so there is no need to seperate it add the logic inside it.  it only exists to be called in one spot.

```
fun saveClientImage(clientId: Int, imageFile: PlatformFile?) {
    imageFile ?: return
    viewModelScope.launch {
        try {
            _showLoading.value = true
            val compressed = compressImage(imageFile, clientId)
            uploadImage(clientId, compressed)
        } catch (e: Exception) {
            _showLoading.value = false
            _clientDetailsUiState.value = ClientDetailsUiState.ShowError(e.message ?: "Unexpected error")
        }
    }
}
```

biplab1 (approved):
Verified the changes, all good to proceed with merging.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientDetails/ClientDetailsViewModel.kt
```diff
@@ -86,8 +80,8 @@ class ClientDetailsViewModel(
                 is DataState.Success -> {
                     _clientDetailsUiState.value = ClientDetailsUiState.ShowUploadImageSuccessfully(
                         result.data,
-                        imageFile.absolutePath(),
                     )
+                    getUserProfile()
```
HekmatullahAmin, line 84: - `getImage()` function from clientDetailsRepo is already returning a Flow<DataState<String>>, then using stateIn(...) or collectAsState() to observe it reactively makes a lot more sense than manually calling it inside `viewModelScope.launch`
- have this in mind if use these types of call always do it like this it is also good for testing and also you won't need to call again for example in `uploadImage` i think you again calling it to notify user. but using above it will directly handle it will listen and update the ui when it is collected in the screen. 
- change the below provided code accordingly
```
val profileImage: StateFlow<ByteArray?> = clientDetailsRepo
    .getImage(clientId.value)
    .map { result ->
        when (result) {
            is DataState.Success -> imageToByteArray(result.data)
            else -> null
        }
    }
    .stateIn(
        scope = viewModelScope,
        started = SharingStarted.WhileSubscribed(5_000),
        initialValue = null
    )
```

HekmatullahAmin (changes requested):

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientDetails/ClientDetailsViewModel.kt
revanthkumarJ: done

core/domain/src/commonMain/kotlin/com/mifos/core/domain/useCases/UploadClientImageUseCase.kt
revanthkumarJ: done

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientDetails/ClientDetailsViewModel.kt
revanthkumarJ: done

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientDetails/ClientDetailsViewModel.kt
revanthkumarJ: moved the discussion to slack

HekmatullahAmin (approved):

niyajali (approved):

PR #2425, opened 2025-07-01
feat: Refactor SignatureScreen to CommonMain using Compose-Signature Library

revanthkumarJ:
Fixes - [Jira-#482](https://mifosforge.jira.com/browse/MIFOSAC-482)

https://github.com/user-attachments/assets/1aff7802-520c-4aaa-b475-5aae81a41ac8


revanthkumarJ:
As i created banch from another prs branch those changes are also appearing here,
Those who are reviewing don't consider those as once that branch merged they won't reflect here
see the files of client signature only for reviewing

Nagarjuna0033 (approved):

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientSignature/SignatureViewModel.kt
```diff
@@ -38,21 +39,31 @@ class SignatureViewModel(
...
+                    _signatureUiState.value =
+                        SignatureUiState.SignatureUploadedSuccessfully
+            }
         }
     }
+
+    fun retry() {
+        _signatureUiState.value =
+            SignatureUiState.Initial
+    }
```
itsPronay: uiState is Error when we get an error from the server, possibly because of internet connection, so when the user presses the retry button, the app should retry uploading the signature to the server again instead of this

for `documentFile == null` you can use Empty UI state or another if needed or don't proceed to this step at all if documentFile is null and show snackbarerror

revanthkumarJ: ok i will refactor and update you

revanthkumarJ: @itsPronay updated can you check once

https://github.com/user-attachments/assets/73ac6b66-797b-45ee-a256-fcaef5e107ac

biplab1:
@revanthkumarJ For official or professional documents, signatures should always be in dark-colored ink (preferably black) on a white background, and this should remain the same for both dark and light themes.

feature/client/src/commonMain/composeResources/values/strings.xml
```diff
@@ -98,6 +98,7 @@
     <string name="feature_client_identifier_isActive">Active</string>
 
     <string name="feature_client_failed_to_add_signature">Failed To Add Signature</string>
+    <string name="feature_client_signature_empty">Signature Should not be empty</string>
```
biplab1: Please align the letter casing in these strings to keep it consistent — for example, use Title Case or Sentence case uniformly (for example: “Signature Should not be empty” vs. “Signature uploaded successfully”, “Failed To Load Identifiers” vs. “Failed to fetch datatable”).

biplab1:
> @revanthkumarJ For official or professional documents, signatures should always be in dark-colored ink (preferably black) on a white background, and this should remain the same for both dark and light themes.

You can configure this directly using `SignatureConfig.Professional` as specified in the docs.

References: 
* [Predefined Configurations](https://github.com/niyajali/compose-signature?tab=readme-ov-file#predefined-configurations)
* [ComposeSign](https://github.com/niyajali/compose-signature?tab=readme-ov-file#composesign)

revanthkumarJ:
@biplab1 yes you are correct the signature should be white background and signature black in all cases, i have updated the code to achieve the functionality. Look at it once and approve if everything is ok.

Note:
Due to some issues in library we can't use SignatureConfig.Professional or Setting background color white , confirmed with niyaj also he told it was a known error and will be resolved soon.

So I achieveied the required functionality differently by wraping ComposeSign with Box.
Later during implementing new ui/ux screens , by that time niyaj brother will resolve the library issue we can remove that box

Current screen in darkmode:
image

itsPronay (approved):

niyajali (approved):

PR #458, opened 2016-12-21
fix: Crash on selecting Centers screen from Nav Drawer

tarun0:
Fix #457 
Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Apply the `MifosStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #2428, opened 2025-07-04
fix(feature:client): fix stuck loading & empty state for charges UI

biplab1:
Fixes - [Jira-#459](https://mifosforge.jira.com/browse/MIFOSAC-459)

Summary of changes:
 - Added `@Serializable` to all charge-related data classes to enable proper serialization/deserialization in API communication.
 - Fix stuck loading state by switching from `collect` to `first()` and handling empty DB results with a proper empty state UI
 - Currency formatting has been added based on the implementation in mobile-wallet.
 - Updated date format to display as `1 July 2025` instead of `2025-7-1`
 
[client-charges-loading-fix.webm](https://github.com/user-attachments/assets/4c111001-8a7f-41df-87c3-70f3ce0a7ad7)

Charges List:
  - Displaying charges list fetched directly from API response (bypassing database save and read)
  
[client-charges-direct-api-fetch.webm](https://github.com/user-attachments/assets/0cebdc44-8fde-4519-92a5-b48de3693eca)

- [x] Run the static analysis check `./gradlew check` or `ci-prepush.sh` to make sure you didn't break anything

- [ ] If you have multiple commits please combine them into one commit by squashing them.

PR #2430, opened 2025-07-07
refactor(feature:client): merge ViewModels, add charge validation, fix empty chargeOptions crash

biplab1:
Fixes - [Jira-#458](https://mifosforge.jira.com/browse/MIFOSAC-458)

Summary of Changes:
   - Fix crash by adding default selected charge values to `ChargeDialogUiState.AllChargesV2` when `chargeOptions` is empty.
   - Merge two dialog view model into client charge view model
   - Show snackbar message on successful charge creation
   - Add input validation dialog input fields and enable/disable button accordingly
   - Make 'ChargeOptions` fields nullable to handle missing field in API response

Before Fix (without charges setup in web-app): 

https://github.com/user-attachments/assets/a06e452b-39d6-4eef-88f0-22263f6f05fd

Before Fix (with charges setup in web-app and no database bypass):

[before-fix-client-charges-dialog-no-charge-setup-no-database-bypass.webm](https://github.com/user-attachments/assets/adaccf54-391c-4aa3-b550-2af57be32a5a)

After Fix:
Note:
 - The newly created charge appears second from the top in the screen recording
 - The charges list is fetched directly from API, bypassing the database 
    (the database handling is a work in progress in another PR)
 
https://github.com/user-attachments/assets/ea30f7e4-0b3a-47a2-9ee4-cacf20b2e48b


biplab1:
**Note**: Merge this after https://github.com/openMF/android-client/pull/2428

Nagarjuna0033:
@biplab1 I think for now we can focus only on the issues that are crashing the application or any api error handling things, because `Show snackbar message on successful charge creation`, `Refactor dialog screen content for better readability and reuse`,
`Add resetUiState to handle dialog state after successful submission` are not that much important as the feature is working. As you know we will be getting new mockups you can take a look into those. we may need to refactor lot of things and may be need to remove current ui components and if necessary, screens also. Then this work will be waste of time. So, it would be better we can fastly wrap up cleaning kmp-impl branch and merging into dev branch. correct me if I am wrong.

@niyajali @revanthkumarJ @itsPronay your thoughts on this

revanthkumarJ:
> @biplab1 I think for now we can focus only on the issues that are crashing the application or any api error handling things, because `Show snackbar message on successful charge creation`, `Refactor dialog screen content for better readability and reuse`, `Add resetUiState to handle dialog state after successful submission` are not that much important as the feature is working. As you know we will be getting new mockups you can take a look into those. we may need to refactor lot of things and may be need to remove current ui components and if necessary, screens also. Then this work will be waste of time. So, it would be better we can fastly wrap up cleaning kmp-impl branch and merging into dev branch. correct me if I am wrong.
> 
> @niyajali @revanthkumarJ @itsPronay your thoughts on this

I also think same like when i see mifos-mobile or template we have used Actions , Events , BaseViewModel and some neat architecture
if we are going to implement similarly here in android client while implementing mockups then, 
 how we handled snackbars , states everything should be moved to viewmodel and change current implementation completely, so may be we can fix small small ui fixes there

biplab1:
@Nagarjuna0033 @revanthkumarJ Just to clarify — the feature wasn’t fully working before.

It was crashing when `chargeOptions` came back empty. I added the necessary `chargeOptions` data so the API response works as expected, and updated the dialog to handle nullable entries to prevent it getting stuck on load.

So the snackbar, state reset, and dialog refactor are not just polish — they fix real flow blockers and ensure the screen stays stable as we transition to the new UI.

Skipping these now risks the same crash or broken flow later.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesViewModel.kt
```diff
@@ -25,20 +25,22 @@ class ClientChargesViewModel(
...
 
-    // for refresh feature
+    private val _clientChargesUiState =
+        MutableStateFlow<ClientChargeUiState>(ClientChargeUiState.Loading)
+    val clientChargesUiState = _clientChargesUiState.asStateFlow()
     private val _isRefreshing = MutableStateFlow(false)
     val isRefreshing = _isRefreshing.asStateFlow()
 
+    init {
+        loadCharges(clientId = clientId.value)
```
itsPronay: We don't need to pass `clientId.value` within the function since its already a local variable. we can do this instead
```
init {
        loadCharges()
}

fun loadCharges() = viewModelScope.launch {
        val response = repository.getClientCharges(clientId.value)
        _clientChargesUiState.value = ClientChargeUiState.ChargesList(response)
    }
```

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientChargeDialog/ChargeDialogScreen.kt
```diff
@@ -163,143 +220,103 @@ internal fun ChargeDialogScreen(
...
+                    chargeId = chargeId,
+                    dueDate = formatDate(dueDate),
+                )
+                onCreate(payload)
+            },
+            modifier = Modifier
+                .fillMaxWidth()
+                .height(50.dp),
+        ) {
+            Text(text = stringResource(Res.string.feature_client_charge_submit))
```
itsPronay: logical error in `isFormValid`, you are setting its value only during initialisation and it is not updated afterwards

itsPronay (changes requested):

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientChargeDialog/ChargeDialogScreen.kt
```diff
@@ -109,14 +107,65 @@ internal fun ChargeDialogScreen(
...
+                    is ChargeDialogUiState.Error -> MifosSweetError(
+                        message = stringResource(state.message),
+                    ) {
+                        onRetry()
+                    }
+
+                    is ChargeDialogUiState.Loading -> MifosCircularProgress()
+
+                    is ChargeDialogUiState.ChargesCreatedSuccessfully -> {
+                        onCreated()
```
itsPronay: image
lets not remove its original implementation, show a snackbar or dialog message as well

biplab1: The snackbar message wasn’t appearing because the dialog screen's success state was immediately replaced by the main screen’s loading state. To fix this, the responsibility of showing the snackbar message has been moved to the Charges screen. You can see the success message in the screen recording above. For more details, refer to: https://github.com/openMF/android-client/pull/2424

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientChargeDialog/ChargeDialogScreen.kt
biplab1: There’s no logical error here. In Jetpack Compose, local variables like `isFormValid` are recalculated on every recomposition, so the UI always reflects the current state correctly.

itsPronay: thanks for the explanation, i was unaware of it

itsPronay: @biplab1 I have a doubt , if these functions were defined outside the composable function instead and acted as independent functions like this

```
fun validateAmount(amount: String): Boolean {
    return amount.isNotEmpty()
}

fun validateChargeName(chargeName: String): Boolean {
    return chargeName.isNotEmpty()
}
```
would the value of `isFormValid` still update automatically on every change? no right?

biplab1: Good question — it will still update automatically. In Compose, any expression inside a composable runs again on recomposition, so `isFormValid` stays up to date as long as you call the validation with the latest state. It doesn’t matter if the functions are defined outside.

It won’t update automatically if you store `isFormValid` in a `remember` state and don’t update it whenever the inputs change.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesViewModel.kt
biplab1: Thanks for the suggestion.

I agree that reading `clientId` directly from `SavedStateHandle` works for the current flow, since the `ViewModel` is recreated for each client and the ID does not change while the instance is alive.

However, I prefer to keep `clientId` explicit in `loadCharges()` for clarity and future-proofing. Passing it makes the data flow obvious, avoids hidden dependencies, and ensures this stays correct if we ever reuse the `ViewModel` or handle `clientId` updates dynamically.

I’d like to keep the parameter for now, but open to discuss if you see any strong reason to remove it.

niyajali:
@Nagarjuna0033 @revanthkumarJ @itsPronay @Aditya3815 @HekmatullahAmin  Guys review this

revanthkumarJ (approved):

itsPronay (approved):

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
```diff
@@ -67,25 +71,28 @@ import org.koin.compose.viewmodel.koinViewModel
...
+    val clientId by clientChargesViewModel.clientId.collectAsStateWithLifecycle()
+    val clientChargeUiState by clientChargesViewModel.clientChargesUiState.collectAsStateWithLifecycle()
+    val refreshState by clientChargesViewModel.isRefreshing.collectAsStateWithLifecycle()
 
     ClientChargesScreen(
         clientId = clientId,
         state = clientChargeUiState,
         onBackPressed = onBackPressed,
-        onRetry = { viewModel.loadCharges(clientId) },
-        onRefresh = { viewModel.refreshCenterList(clientId) },
```
niyajali, line 85: Call by method reference(::)

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
```diff
@@ -67,25 +71,28 @@ import org.koin.compose.viewmodel.koinViewModel
...
         refreshState = refreshState,
         onChargeCreated = {
-            viewModel.loadCharges(clientId)
+            // resetUiState() is needed here to clear the success state immediately after successful
+            // client charge creation, so that reopening the dialog doesn’t reuse stale state and
+            // accidentally retrigger main screen loading.
+            // Downside: this causes two back-to-back loading states — one in the dialog and one on
+            // the main screen.
+            chargeDialogViewModel.resetUiState()
+            clientChargesViewModel.loadCharges(clientId)
```
niyajali: Why there are multiple action do whatever needs inside of the view model

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
```diff
@@ -67,25 +71,28 @@ import org.koin.compose.viewmodel.koinViewModel
 @Composable
 internal fun ClientChargesScreen(
     onBackPressed: () -> Unit,
-    viewModel: ClientChargesViewModel = koinViewModel(),
+    clientChargesViewModel: ClientChargesViewModel = koinViewModel(),
```
niyajali: Why there are multiple ViewModel fix this, these are anti-pattern

niyajali (changes requested):

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
biplab1: These two calls handle separate concerns: `resetUiState()` clears the dialog’s transient success state to avoid stale reuse, while `loadCharges()` refreshes the main screen list to reflect the new charge. Both are needed together to keep the UI consistent. And these are actions on two separate view models.

I thought about putting the `resetUiState()` inside `ChargeDialogViewModel` too, but then the success state disappears before the snackbar or any UI feedback can happen. Adding a delay just to keep it alive isn’t reliable.

The core issue is that after a successful submission, the dialog stays stuck on `ChargeCreatedSuccessfully`, so reopening it too soon can reuse stale state and retrigger `onChargeCreated()` unnecessarily.

Calling `resetUiState()` and `loadCharges()` together right after creation is the cleanest way for now — it clears the stale state of dialog and refreshes the main screen list. Only downside of this approach is back-to-back loading states — one in dialog and another in main screen. For more details, refer to https://github.com/openMF/android-client/pull/2424 where we discussed this earlier.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
biplab1: The concern about multiple ViewModels being an anti-pattern doesn’t apply here. Google’s *Now in Android* example shows a clear case where using more than one ViewModel per screen is valid — see [InterestsListDetailScreen](https://github.com/android/nowinandroid/blob/main/app/src/main/kotlin/com/google/samples/apps/nowinandroid/ui/interests2pane/InterestsListDetailScreen.kt).

In that screen, they use:

* `Interests2PaneViewModel` → manages overall coordination and selection state
* `TopicViewModel` → handles topic-specific details with its own lifecycle

My use case is similar:

* `ChargeDialogViewModel` → manages charges template and dialog-specific state 
* `ClientChargesViewModel` → manages the main charges list and screen state

This separation follows single responsibility and keeps logic testable and scoped. The brief overlap (calling `resetUiState()` then `loadCharges()`) is just coordination between related but separate concerns.

It `would` be an anti-pattern if the ViewModels duplicated the same logic, shared mutable state carelessly, or tightly coupled flows that should be independent. But here each ViewModel owns distinct, isolated state.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
biplab1: I’d like to keep `clientId` explicit even if it’s technically redundant for now (discussed in an earlier comment [here](https://github.com/openMF/android-client/pull/2430#discussion_r2201665850)). It keeps the flow clear and makes `loadCharges()` pure and easier to test or adapt later if we ever support dynamic client switching.

Keeping the parameter does mean we can’t switch to `::loadCharges` because the method reference can’t bind an argument — so the current `{}` lambda is the right approach here.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientChargeDialog/ChargeDialogScreen.kt
HekmatullahAmin: @biplab1 I just saw your chat with @itsPronay  and then saw validateAmount and validateChargeName. why we have functions inside composables.

- These functions are only used to check whether a field is empty and to update local error states. Since they directly reference and modify local state (amountError, chargeNameError), they were placed inside the Composable. But honestly — defining functions like that inside a Composable isn’t ideal.

- Instead of creating separate validate...() functions, it's cleaner and more readable to do validation inline where it's used. For example:(just an exmpale of inline validation)
```
amountError = amount.isEmpty()
chargeNameError = chargeName.isEmpty()
val isFormValid = amount.isNotEmpty() && chargeName.isNotEmpty() 
```


- If the button is already conditionally disabled using:
```
enabled = isFormValid
```
then the onClick handler should not need to validate again or return early using return@Button. That logic becomes redundant, because:
* If the form is invalid, the button can't be clicked.
* If the button is enabled, it means the form has already passed basic validation.
So checking again in onClick and using return@Button is extra boilerplate that can be safely removed unless something else is happening which u handle it accordingly. the thing is we should make the code readable for others everyone sees it gets confused.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
HekmatullahAmin: Thanks for the detailed explanation, @biplab1  — I appreciate your perspective.
- You can do that, of course, but in this context it’s not really necessary. Since the ViewModel is scoped to a single client and the `clientId` is already available via `SavedStateHandle`, passing it explicitly doesn’t add much value here. In fact, by reading it directly, we could simplify the code and also use a method reference like `::loadCharges`. 
- Regarding testability, we can still maintain clean and testable code without the parameter, as the repository can be mocked, and the behavior of `loadCharges()` can be easily validated.  
- As for dynamic client switching — I honestly don’t think that use case will come up. This screen is meant to display charges for a specific client, passed during navigation, and that client ID remains fixed throughout the ViewModel’s lifecycle. 
Just wanted to share my reasoning on why simplifying it might be worthwhile too.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
HekmatullahAmin: @biplab1 , I don’t think the comparison is fully equivalent to our case. 

- In `Now in Android`, they use assisted injection (via a factory) to pass the topic ID into the TopicViewModel. In our project, we typically use navigation arguments and retrieve them through SavedStateHandle. cause of that they used it there. and it will only be initialized of the one topic clicked.  
- In our case, you including both ClientChargesViewModel and ChargeDialogViewModel in the same screen, even though the dialog is not always visible. This means both ViewModels get initialized upfront — even when the user hasn’t opened the dialog.  
- If it were up to me, I would personally prefer keeping the dialog logic within the main ViewModel — especially since the dialog is relatively temporary and belongs to this screen. That would simplify lifecycle handling and reduce coordination (like needing to call resetUiState() manually).

- As I mentioned in your last PR, adding a comment would help future contributors understand why this decision was made. but mentioned, this is something we should aim to simplify later.

Just sharing my view

niyajali: The reference you provided doesn't demonstrate multiple ViewModel injections on a single screen - it's actually a NavigationPaneScaffold implementation with animations. When users click on a topic, the topic details animate into view on top of the scaffold rather than navigating to a new screen. They use a Factory class to pass the topic ID to savedStateHandle because while the topic screen can be navigated to from other screens (requiring navigation), in this specific case it's simply animating the content display without actual navigation.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
niyajali: I'm expressing the same point

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
biplab1: @HekmatullahAmin 
* My point with the Now in Android example wasn’t that our setup is identical — just that using multiple ViewModels in a single screen can be valid when there’s clear separation of concerns. So it’s similar in principle, not exactly equivalent.
* You’re right — in our case, the `ChargeDialogViewModel` is initialized upfront, even if the dialog stays closed. With the current flow, this was the cleanest way to handle the transient success state without stale state issues.
* I agree that if we moved the dialog logic into `ClientChargesViewModel`, we wouldn’t need the extra `resetUiState()` at all — the flow would naturally clear its own state without extra coordination. So the extra call is only needed because the dialog and main logic are separate right now.
* A comment explaining this is already in the code, but I can expand it if you think it’s not clear enough for future maintainers.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
biplab1: @HekmatullahAmin @niyajali @itsPronay Fair point — I’ll simplify it and switch to reading `clientId` directly so we can use `::loadCharges`. Thanks for the clear feedback.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientChargeDialog/ChargeDialogScreen.kt
biplab1: Got it, I’ll update it to inline validation as suggested — thanks!

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
biplab1: @niyajali You’re right — Now in Android uses `NavigationPaneScaffold` and lazy-loads the `TopicViewModel` only when the detail pane is shown, passing the topic ID with an assisted factory. So technically it’s not injecting both ViewModels upfront.

My point with that reference wasn’t about how the injection or pane scaffold works, but that it shows a valid pattern: using more than one ViewModel in the same screen when parts of the UI — like a persistent list and a transient detail — have clearly separate responsibilities and lifecycles.

In our case, the charge dialog works similarly in that it overlays the main list without full navigation, so the split ViewModels keep concerns separate. The main difference is that our `ChargeDialogViewModel` is always initialized upfront — that’s something we can revisit to scope better later.

niyajali: @biplab1 Yes, multiple ViewModels can be used in certain scenarios, but this isn't typically necessary. If dialog state needs to be shared/sharing across multiple screens, consider creating a base class like BaseViewModel that your ChargeViewModel can inherit from. This approach allows easy access to dialog state while maintaining a single ViewModel instance rather than multiple ones. Using multiple ViewModels has several drawbacks: they're bound to the activity lifecycle and must be properly cleared when screens close. If a ViewModel isn't cleared correctly, it can cause memory leaks based on my experience.

For more information about ViewModel lifecycle and best practices, visit the official Android developer guide.

biplab1: @niyajali Thanks for the discussion — I think we’ve covered all angles on this. To avoid prolonging it further, I’ll revert back to the original approach without `resetUiState()`, which also removes the need for `ChargeDialogViewModel` dependency injection in the main screen.

This does mean the main screen may reload again if someone clicks the + icon right after a successful submission, but that’s an acceptable tradeoff for now.

If needed, we can revisit this flow later and decide whether to merge the dialog logic or adjust the scoping. For now I’ll align with your preference and simplify it.

https://github.com/user-attachments/assets/51cd462c-e267-45d3-9ae3-5021153dd13a

niyajali: Why don't you just move the dialog and other states logic to a single ViewModel as the dialog meant to designed for this screen and when user created new charge or edited, deleted show the snackbar and reset the dialog fields state if needed.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientChargeDialog/ChargeDialogScreen.kt
biplab1: @HekmatullahAmin Thanks for pointing this out — you’re right, having local functions inside the Composable wasn’t ideal. I’ve updated it so that `validateAmount` and `validateChargeName` are now pure functions outside the Composable.

`validateAmount` now matches the Mifos web app logic: it checks not just for empty input but also for valid number format and correct decimal places. It returns an enum (`EMPTY`, `INVALID_FORMAT`, `VALID`) so the UI can show either “Field required” or “Invalid amount format” clearly.

The `Button` stays disabled unless the form is valid, so the onClick only marks fields as touched to show errors — no redundant `return@Button`.

Let me know if you’d like any more adjustments.

https://github.com/user-attachments/assets/efaf48e7-a853-4ccc-8a34-58119e37f74e

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientCharges/ClientChargesScreen.kt
niyajali: Are we're using same dialog on multiple cases by changing the dialog content or what I'm not understanding by seeing the above video it's like simple though.

biplab1: @niyajali  Just for context, the same earlier approach which I reverted (separate dialog ViewModel + `resetUiState()` to clear stale success state) was used in the previous *Identifiers* PR and got merged with the same tradeoff.

I kept the same pattern here to stay consistent. But if you prefer, I’m fine merging the dialog logic into the main screen ViewModel instead.

Could you please clarify your last sentence — are you asking if this dialog is reused in multiple places with different content? If so, no — it’s just for creating a charge in this one screen.

niyajali: @biplab1 I usually approve and merge PRs without reviewing them myself, trusting that other team members' approvals mean the implementation is solid. 

And yes that's the question in that if it's not used in other screens we won't need any ViewModel for that, move all states to a single ViewModel and update the codebases accordingly.

If you want to share or reuse state, consider creating an open or abstract class and inheriting from it—this achieves the same result. However, ViewModel is different because it's bound to the lifecycle and can cause memory leaks if not properly cleared

If you've implemented similar patterns with multiple ViewModels elsewhere, please address those issues first after this PR.

niyajali: @revanthkumarJ @itsPronay Are we following this review process for PRs? And We should focus on sustainable solutions rather than quick fixes.

biplab1: @niyajali Got it — thanks for clarifying.

I agree that merging the dialog logic into the main ViewModel makes sense for this screen since it’s not reused anywhere else. I’ll update this PR to do that and also check the Identifiers flow and any similar patterns to align them with this approach.

Just to clarify: the dialog ViewModel DI + `resetUiState()` (in main screen) was intentional to avoid stale state bugs and coordinate the dialog success -> reload flow. It wasn’t meant as a quick workaround — but I agree that merging keeps it simpler overall for a single-use dialog.

I’ll clean this up and push the updated version soon.

biplab1:
@niyajali I’ve merged the two view models as discussed. Please let me know if anything else needs to be done. If everything looks good, kindly go ahead and merge it.

niyajali (approved):

PR #464, opened 2017-01-02
Solves overlapping of frags in DashboardActivity

SKrPl:
solves this issue: https://github.com/openMF/android-client/issues/397

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Apply the `MifosStyle.xml` style template to your code in Android Studio.

- [ ] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #2433, opened 2025-07-08
feat(feature:client): implement Pinpoint Map dialog and fix address flows

biplab1:
Fixes - [Jira-#464](https://mifosforge.jira.com/browse/MIFOSAC-464)

Summary of Changes:
 - Implement Pinpoint Map as a dialog to add/update address
 - Fix add, update, and delete address flows for PinpointClientScreen
 - Add proper location permission checks and fallback handling for user location
 
**Note:**
Android-only for now — permissions are covered with `PermissionBox`. Will refactor to `commonMain` and explore KMP map options in a separate PR.

Before Fix:

[before-fix-client-pinpoint.webm](https://github.com/user-attachments/assets/494e6e04-da7d-422c-88d3-8a9f328968bb)

After Fix:

https://github.com/user-attachments/assets/240f2a28-8136-479b-adae-da79d74e90ef


revanthkumarJ:
Mentioning these libraries for reference , if it aligns with our usecase use them as they support multi platform or continue with current approach which works on only android

@biplab1 

for permissions we can use moko-permissions libarry 

https://github.com/icerockdev/moko-permissions

and for maps we can use

https://github.com/[farimarwat/KrossMap](https://github.com/farimarwat/KrossMap)

see them once if they align with this use them otherwise continue with current approach.

revanthkumarJ:
@biplab1 see them once if they don't align with use case ping me i will approve the current one , it looks good to me for android implementation

biplab1:
@revanthkumarJ Thanks for sharing these!

This implementation is Android-only for now, since we needed a working version first. Once that’s stable, I’ll refactor it to `commonMain` for true multiplatform support.

* **Permissions:** We already handle this with our `PermissionBox`, so no extra library is needed.
* **Maps:** `KrossMap` only supports Android and iOS and isn’t widely used yet, so I’d prefer to stick with our current approach for now. The Android part looks big mainly because the map dialog is inline — I’ll split that into separate files to keep it clean.

I’ll keep looking for better KMP options for map.

revanthkumarJ (approved):

niyajali (approved):

PR #2435, opened 2025-07-10
fix: Update server config not working 

itsPronay:
Fixes - https://mifosforge.jira.com/browse/MIFOSAC-494

https://github.com/user-attachments/assets/99513898-801d-44ab-a7ea-e4037e2ff9f1


feature/settings/src/commonMain/kotlin/com/mifos/feature/settings/updateServer/UpdateServerConfigViewModel.kt
```diff
@@ -119,6 +119,7 @@ class UpdateServerConfigViewModel(
                     ).any { it.value != null }
 
                     if (!hasAnyError) {
+                        prefManager.logOut()
```
HekmatullahAmin, line 122: are u sure to clean everything from `userpreference` in this operation?

revanthkumarJ (approved):
Looks good to me ,

biplab1 (approved):
The changes are fine from my side, good to go.

core/network/src/commonMain/kotlin/com/mifos/core/network/di/NetworkModule.kt
```diff
@@ -38,10 +39,16 @@ val NetworkModule = module {
         }
     }
 
+    single<String>(named("baseUrl")) {
+        val preferencesRepository = get<UserPreferencesRepository>()
```
HekmatullahAmin, line 42: move the `named("baseUrl")` in `Qualifier.kt` file and here directly call like it is done for `MifosClient`

feature/settings/src/commonMain/kotlin/com/mifos/feature/settings/updateServer/UpdateServerConfigViewModel.kt
HekmatullahAmin: Calling `prefManager.logOut()` removes all saved values from persistent storage — this includes:

UserData, UserInfo, ServerConfig, AppTheme etc. 
I believe you restart the app cause u updated the url or whatever, but does it mean the user should also lose his other states like selected app theme. I don't think it would be ideal for doing one operation also done in back sth that user didn't ask for. ask @niyajali about it.

itsPronay: logout should not clear appThemes, I am not sure why its there. Maybe whoever did the `UserPreferenceRepository` took reference from `mobile wallet` since we don't have apptheme in there. 
 
 thanks for pointing it out @HekmatullahAmin

core/datastore/src/commonMain/kotlin/com/mifos/core/datastore/UserPreferencesDataSource.kt
```diff
@@ -142,7 +133,7 @@ class UserPreferencesDataSource(
 
     suspend fun clearInfo() {
         withContext(dispatcher) {
-            settings.clear()
+            settings.putAuth(User())
```
niyajali, line 136: Where does this method defined, and just clear the AUTh_USER and other related keys and keep the app specific configured as it is while logging out

niyajali (changes requested):

core/datastore/src/commonMain/kotlin/com/mifos/core/datastore/UserPreferencesDataSource.kt
itsPronay: @niyajali  `clearInfo()` is used while logging out and before updating server config.
I tried replacing it with this block of code, but it doesn't work correctly
```
    suspend fun clearInfo() {
        withContext(dispatcher) {
            settings.remove(AUTH_USER)
        }
    }
```
   so I had to use `settings.putAuth(User())`

niyajali:
@itsPronay please upload a video to demonstrate the changes

itsPronay:
@niyajali I have uploaded the video, please check

niyajali (approved):

PR #2437, opened 2025-07-16
fix(feature:path-tracking): fix location permission handling and correct package name

biplab1:
Fixes - [Jira-#495](https://mifosforge.jira.com/browse/MIFOSAC-495)

Summary of changes:
 - Fixed user location permission handling
 - Corrected the package name 
 
Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Run the static analysis check `./gradlew check` or `ci-prepush.sh` to make sure you didn't break anything

- [ ] If you have multiple commits please combine them into one commit by squashing them.

PR #2441, opened 2025-07-27
fix(feature:path-tracking, core:network): fix bugs in path tracker fe…

sam-arth07:
Fixes - [Jira-#495](https://mifosforge.jira.com/browse/MIFOSAC-495)

Before:

After: 

https://github.com/user-attachments/assets/74dcb4a7-c3b6-4f42-a4a1-af243913ff24 
 
### Changes made in the PR: 
- Included changes related to `location permission handling and correct package name` already done in this draft pr: https://github.com/openMF/android-client/pull/2437 
- resolved 404 network calls by creating a new datatable `m_staff_path_tracking ` with appropriate fields and updated the `DataTableService` to use this new datatable endpoint. 
- Added Logic to display accurate start, stop, and route markers in `PathTrackingScreen.android.kt`.
- Validated that timestamps match the actual tracking session and multiple sessions display correctly.
- Verified that the functionality is working on different network conditions (4G (LTE), 3G (UMTS) , 2G (EDGE), WiFi).


core/model/src/commonMain/kotlin/com/mifos/core/model/objects/users/UserLocation.kt
```diff
@@ -11,10 +11,12 @@ package com.mifos.core.model.objects.users
 
 import com.mifos.core.model.utils.Parcelable
 import com.mifos.core.model.utils.Parcelize
+import kotlinx.serialization.Serializable
 
 /**
  * Created by Rajan Maurya on 24/01/17.
  */
+@Serializable
```
therajanmaurya, line 19: We are no longer using Serializable, @niyajali confirm?

therajanmaurya (dismissed):

.gitignore
```diff
@@ -12,4 +12,5 @@ mifosng-android/mifosng-android.iml
 *.hprof
 android-client.iml
 /secrets.properties
-.kotlin/
\ No newline at end of file
+.kotlin/
+build-logic/convention/bin
```
HekmatullahAmin: use `bin/` so all of them git ignored

feature/path-tracking/src/androidMain/kotlin/com/mifos/feature/path/tracking/PathTrackingScreen.android.kt
```diff
@@ -127,35 +124,36 @@ actual fun PathTrackingMapView(latLngList: List<UserLatLng>) {
...
     GoogleMap(
         modifier = Modifier
             .fillMaxWidth()
             .height(150.dp),
         cameraPositionState = cameraPositionState,
         uiSettings = uiSettings,
-    )
-}
+    ) {
+        if (latLngList.isNotEmpty()) {
```
HekmatullahAmin, line 136: use proper name for `latLngList` it is not understandable. maybe use full form instead of abbreviation

core/model/src/commonMain/kotlin/com/mifos/core/model/objects/users/UserLocation.kt
HekmatullahAmin: @therajanmaurya i belive for now we are no longer using parcelizable cause it is android specific, and in ios it doesn't have some annotations

core/network/src/commonMain/kotlin/com/mifos/core/network/datamanager/DataManagerDataTable.kt
```diff
@@ -71,7 +71,7 @@ class DataManagerDataTable(
     }
 
     /**
-     * This Method is adding the User Tracking Data in the data Table "user_tracking"
+     * This Method is adding the User Tracking Data in the data Table "client_pinpoint_location"
```
biplab1: It should be `m_staff_path_tracking` instead.

core/network/src/commonMain/kotlin/com/mifos/core/network/datamanager/DataManagerDataTable.kt
```diff
@@ -85,7 +85,7 @@ class DataManagerDataTable(
     }
 
     /**
-     * This Method is fetching the User Path Tracking from the DataTable "user_tracking"
+     * This Method is fetching the User Path Tracking from the DataTable "client_pinpoint_location"
```
biplab1: Same here, please review

biplab1 (commented):
Thanks for the demo! I have a few questions for better clarity:
1. Could you share a screenshot of the backend data that was captured?
2. Which entity was used to create the DataTable used for path tracking (e.g., client, group, center, office)?
3. Where in the `web-app` can we view the DataTable values for the user shown in the screen recording?
4. What do the start and end positions represent? Are they the actual initial and final locations of the trip? Similarly, are the start time and end time automatically recorded timestamps, or are they estimated durations between manually selected points?
>Just to clarify — the start and end locations should be captured automatically when the journey begins and ends, not based on estimated locations selected by the field officer on the map.
5. The field officer shouldn’t be able to edit start/end locations manually — these should come from live GPS data once tracking begins.
6. How is path tracking started or stopped?
7. Was this tested on a physical device or an emulator?
    (For location-based features, physical devices are preferred as emulators may return default or inaccurate GPS data.)
8. When a field officer starts path tracking and stops it upon reaching the destination, a new card should appear in the UI, and the corresponding data should be saved to the DataTable.
9. Edge cases should be handled — for example, if the field officer starts and stops tracking at nearly the same location (within a small epsilon distance), then that entry should be ignored and not recorded.

feature/path-tracking/src/androidMain/kotlin/com/mifos/feature/path/tracking/PathTrackingScreen.android.kt
biplab1: Just a note — LatLng is actually the standard term used in the Google Maps API, so using latLngList here is appropriate and widely understood in that context.

Example:
`position = CameraPosition.fromLatLngZoom(LatLng(latLng.lat, latLng.lng), 15f)`

sam-arth07:
> Thanks for the demo! I have a few questions for better clarity:
> 
> 1. Could you share a screenshot of the backend data that was captured?
> 2. Which entity was used to create the DataTable used for path tracking (e.g., client, group, center, office)?
> 3. Where in the `web-app` can we view the DataTable values for the user shown in the screen recording?
> 4. What do the start and end positions represent? Are they the actual initial and final locations of the trip? Similarly, are the start time and end time automatically recorded timestamps, or are they estimated durations between manually selected points?
> 
> > Just to clarify — the start and end locations should be captured automatically when the journey begins and ends, not based on estimated locations selected by the field officer on the map.
> 
> 5. The field officer shouldn’t be able to edit start/end locations manually — these should come from live GPS data once tracking begins.
> 6. How is path tracking started or stopped?
> 7. Was this tested on a physical device or an emulator?
>    (For location-based features, physical devices are preferred as emulators may return default or inaccurate GPS data.)
> 8. When a field officer starts path tracking and stops it upon reaching the destination, a new card should appear in the UI, and the corresponding data should be saved to the DataTable.
> 9. Edge cases should be handled — for example, if the field officer starts and stops tracking at nearly the same location (within a small epsilon distance), then that entry should be ignored and not recorded.


@biplab1 

1. 
- image

 - image

2. client entity
3. https://github.com/user-attachments/assets/fe626a22-68e4-4455-9dc3-e7cf2e479cc1

4. Start and end positions are the actual location that is represented by the original (latitude, longitude) and destination (latitude, longitude) resp. No they are not being recorded automatically in the current implementation. I have mocked dummy data in the m_path_tracking datatable by looking at the type of data the frontend expects. 

``` 
onPathTrackingClick = { userLatLngs ->
            val uri = if (userLatLngs.isNotEmpty()) {
                val originLatLng = userLatLngs[0]
                val destinationLatLng = userLatLngs[userLatLngs.size - 1]
                "https://maps.google.com/maps?f=d&hl=en&saddr=${originLatLng.lat},${originLatLng.lng}" +
                    "&daddr=${destinationLatLng.lat},${destinationLatLng.lng}"
            }
```
``` PathTrackingScreen.kt
@Composable
private fun PathTrackingItem(
    pathTracking: UserLocation,
    modifier: Modifier = Modifier,
    onPathTrackingClick: (List<UserLatLng>) -> Unit,
) {
    val latLngList = getLatLngList(pathTracking.latLng)
    OutlinedCard(
        modifier = modifier.padding(8.dp),
        onClick = { onPathTrackingClick(latLngList) },
        colors = CardDefaults.outlinedCardColors(MaterialTheme.colorScheme.surface),
    ) {
        PathTrackingMapView(latLngList = latLngList)

        Text(
            modifier = Modifier.padding(8.dp),
            text = "${pathTracking.date} from ${pathTracking.startTime} to ${pathTracking.stopTime}",
            style = MaterialTheme.typography.bodyLarge,
        )
    }
}

````

5. Ok
6. I had a talk on the slack channel regarding the same, can be viewed in this thread: https://mifos.slack.com/archives/C06FC73DZPW/p1753377376922779

> Timestamps and coordinates need verification.
Test the complete flow: start tracking → record path → stop tracking → display map with route.
How do I start tracking sessions? there seems to be no way to begin a new session, only display the stored sessions from the backend.

Do i need to include the start tracking feature in this PR only or it can be done in a separate PR? 

7. I tested this on both emulator (for simulating slower networks) and physical device for overall testing and the recording is from a physical device.

8. Not yet impl
9. Not yet impl

biplab1:
Here are my thoughts on your responses and the current implementation:

1. I was referring to the data captured at the backend, not the logcat output. From your screen recording (point 3), it appears that no actual data is being captured — only dummy values are present in the `DataTable`.

2. Path tracking data shouldn't be linked to the `client` entity, as it's related to `staff` (field officers). A staff member may or may not be a client of the organization.
   A better approach would be to associate the path tracking `DataTable` with the `office` entity, since staff are linked to offices. You can save both `officeId` and `staffId` in the table.
   To create a staff member, refer to:
   * [Manage Employees](https://docs.mifos.org/mifosx/user-manual/for-administrators-mifos-x-platform/administration/organization/manage-employees)
   * [Create Users](https://docs.mifos.org/mifosx/user-manual/for-administrators-mifos-x-platform/administration/users)
   
     After creating an employee, you'll also need to assign them as a user.

3. Path tracking data should not appear on the client page — there is no such concept as `client` path tracking. Please remove the table from the DataTables section.
   Also, it looks like you're currently fetching the path tracking `DataTable` using `clientId` instead of the ID of the currently logged-in `user`.

4. This indicates that path tracking was never properly implemented. The original version may have been using mocked data as well.

5. This can be handled in the same PR as in point 6

6. It's your choice whether to address these issues in this PR or in a follow-up PR.

7. Thanks for confirming. Good to know the recording is from a physical device — that's important for reliable GPS testing, since emulators can return default values.

8 & 9. You can add the following improvements in your next PR:

* When a field officer starts and then stops path tracking, a new card should appear in the UI, and the values should be stored in the `DataTable`.
* Handle edge cases where tracking starts and stops at nearly the same location (within some epsilon distance) — in such cases, the entry should be ignored to avoid unnecessary records.

10. Optional Suggestion:
If feasible, path tracking could show the actual route taken, similar to how apps like Uber or Ola display ride history — instead of just a straight line between the start and end locations.

     Alternatively, if the goal is only to capture start and end points, a separate user flow could be introduced. In that flow, the app would record only the GPS coordinates at the beginning and end of the journey, manually (by clicking the location icon to fetch current location) — this would significantly reduce battery usage by avoiding continuous location tracking. However, this raises an important question:
     > Would this still be considered "path tracking," or simply a start/end location record?

sam-arth07:
@HekmatullahAmin @niyajali I have made the requested changes, if there is anything else you can let me know, I will try to make changes as earliest as possible, else it is ready to be merged.

Also @biplab1, I have changed the datatable from prev being linked to client entity to a office entity now which is then assigned to a user as well, also deleted the `m_staff_path_tracking` datatable from the client datatables section. 

Attached screenshots for reference:

image

image

As for the start path tracking feature, i had a discussion with @therajanmaurya , he mentioned that it could be taken up in future PRs. For this PR, only being able to display the mapview on the Path Tracking Screen is enough.
Thank you for your suggestions and guidance.

niyajali:
@sam-arth07 Can you store more information like place name of that lat and long using gmap API so we can show in path tracking card instead of time.

sam-arth07:
@niyajali Ok, I have made changes to display the start and endpoint address on the pathTracking card along with Time. If the startAddress and endAddress is not present in the data of the path tracking datatable then we fallback to using the nominatim opensteetmap api to decode the address from the location's lattitude and longitude .

Also updated the `UserLocation` dataClass and the `m_staff_path_tracking` to accomodate for startAddress and endAddress: 

image

---

image

---


Does this look good to you?

HekmatullahAmin (approved):

itsPronay (approved):

PR #2456, opened 2025-08-19
feat : added listing components ui

revanthkumarJ:
Fixes - [Jira-#513](https://mifosforge.jira.com/browse/MIFOSAC-513)

image


core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosListingComponent.kt
```diff
@@ -0,0 +1,657 @@
...
+            }
+        }
+    }
+}
+
+@Composable
+fun MifosChargesListingComponent(
+    title: String,
+    type: String,
+    dueAsOf: String,
```
Nagarjuna0033, line 423: @revanthkumarJ instead of passing all these can we pass Map<String, String> like this if possible

Nagarjuna0033 (changes requested):

core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosListingComponent.kt
revanthkumarJ: one overload is there for that one too.
because of different colours i created different functions

revanthkumarJ: it didn't followed the same structure , in one card 2 items grouped in another 3 grouped and so i created different overloads to use differently

Nagarjuna0033:
```
data class ListingRow(
    val key: String = "",
    val value: String,
    val keyStyle: TextStyle? = null,
    val valueStyle: TextStyle? = null,
    val valueColor: Color? = null,
    val isHeader: Boolean = false,
    val spacerAfter: Dp? = null,
)

```

May be You can create a data class like this and try it out

Issue #583, opened 2017-03-14
Progress bar Starts Rotating Without Reloading fragment


meghalagrawal:
**Summary:** 

When we are in Clients Fragment and opening any item from menu ( ex - Create Client ) and then coming back ( using default Android Back button ) then ProgressBar starts Loading without Reloading or Without Calling API.

Summarise your issue in one sentence (what goes wrong, what did you expect to happen)

**Steps to reproduce:** 

How can we reproduce the issue?

We need to stop this behaviour of Progressbar and the Fragment should not even Reload Needlessly

**Expected behaviour:** 

What did you expect the app to do?
The Clients Fragment should not be reloaded neither the Loader should be shown. The Clients Fragment should be shown from stack.

**Observed behaviour:** 

What did you see instead?  Describe your issue in detail here.

The Loader was shown needlessly when we are clicking back button and coming back to Clients Fragment.

**Device and Android version:** 

Android 6.0.1 ( MI Note 4 )

version from the manufacturer or a custom ROM?

**Screenshots:** 
![screenshot_2017-03-15-03-01-38-533_com mifos mifosxdroid](https://cloud.githubusercontent.com/assets/18400951/23923248/e61ba072-092b-11e7-9be8-dd9659d39b93.png)

meghalagrawal:
This issue was not only for Loader but it was also for RxJava Composite Subscription. Once we were Unsubscribing to it we were unable to Subscribe to it again.

A Better issue is submitted by me here #587 

Pull Request has been added for it Please Review #588

Aditya-gupta99:
This issue has been already resolved.

PR #2476, opened 2025-08-27
feat: Migrate Navigation to Type Safe Navigation

revanthkumarJ:
Fixes - [Jira-#583](https://mifosforge.jira.com/browse/MIFOSAC-583)


niyajali (changes requested):
@revanthkumarJ Copy the `cmp-navigation` module from template project and update the config accordingly as we've discussed about this module on call earlier and you have worked on other project which has same configuration.

niyajali (approved):
@revanthkumarJ build is failing, though I'm approving this fix the build issue and proceed with merge..

PR #588, opened 2017-03-15
fix:#587 Re Subscribe to Observer onAttachView()

meghalagrawal:
After Unsubscribing the Observer it was not able to get Subscribed again to we initialised Composite Subscription again in onAttachView that now enables us to use Swipe to refresh and other things on Reload of the view

The issue #587 is Fixed and is Working

Please make sure these boxes are checked before submitting your pull request - thanks!

- [Not Tick as i didnt Found this Xml ] Apply the `MifosStyle.xml` style template to your code in Android Studio.

- [Tick ] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [Tick ] If you have multiple commits please combine them into one commit by squashing them.

PR #2485, opened 2025-09-01
fix notes screen issue

Arinyadav1:
Fixes - [Jira-#588](https://mifosforge.jira.com/browse/MIFOSAC-588)

It fix the issue note list not update immediately after update and add note.


feature/note/src/commonMain/kotlin/com/mifos/feature/note/addEditNotes/AddEditNoteScreen.kt
```diff
@@ -112,7 +114,12 @@ fun AddEditNoteScreenDialog(
             )
         }
 
+        AddEditNoteState.DialogState.Success -> {
+            onAction(AddEditNoteAction.NavigateBackWithUpdateList)
```
itsPronay: Why are you using `DialogState.success` when you are not showing any dialogsMessages

Arinyadav1: It is not show dialog but when the user click update/add button then after success the api it will navigate to notesList screen that is why that is there.

itsPronay: this should not be here, put it inside the viewmodel and remove success if it's not related to dialog, error or loading

feature/note/src/commonMain/kotlin/com/mifos/feature/note/notes/NoteScreen.kt
```diff
@@ -152,10 +160,12 @@ internal fun NoteScreenScaffold(
...
-                NoteContent(
-                    state = state,
-                    onAction = onAction,
-                )
+                if (state.dialogState != NoteState.DialogState.Loading && !state.isError) {
+                    NoteContent(
+                        state = state,
+                        onAction = onAction,
+                    )
+                }
```
TheKalpeshPawar: is the !state.isError necessary does't dialog state have a error type too.

TheKalpeshPawar (changes requested):

feature/note/src/commonMain/kotlin/com/mifos/feature/note/notes/NoteScreen.kt
Arinyadav1: If I dont this then if the error occur then the behind content is visible with error message it is simply hide the content when error occur.

Arinyadav1: @TheKalpeshPawar 

??

itsPronay: @Arinyadav1 , This logic doesn't fit tbh, see other screeens and try to take a reference

Arinyadav1:
@itsPronay 

what happens ?

itsPronay:
@Arinyadav1 Please make the following changes, see other screens for reference
1. Remove custom Note Empty Screen that you made as we already have this component.  Use `MifosEmptyCard`,
2. Move Note Compoents to `com.mifos.core.ui.components`, we already have similar design, follow its style, no need to start from scratch, it just gives extra code.
3. This should go inside the dialogState, all the dialog related things go inside dialogState, we don;t need another state variable for this
image

Arinyadav1:
@itsPronay 
`MifosEmptyCard,` does not have `shape` `elevation` then how can I use it instead custom Note Empty Screen

itsPronay:
shape ? elevation?  I don't see Empty screen having shave and elevation in figma screen.  why do you need it when every empty screen looks the same.

feature/note/src/commonMain/kotlin/com/mifos/feature/note/notes/NoteScreen.kt
```diff
@@ -152,10 +159,12 @@ internal fun NoteScreenScaffold(
...
```
TheKalpeshPawar: sorry I wasn't available.
I saw your viewmodle for notescreen and you have isError as an extra flag You are only updating its value to true when you get DialogState.Error and false for other cases. It's seems like an unnecessary flag.

TheKalpeshPawar (changes requested):

feature/note/src/commonMain/kotlin/com/mifos/feature/note/notes/NoteScreen.kt
TheKalpeshPawar: Take a look at this code:

```kotlin
@OptIn(ExperimentalMaterial3Api::class)
@Composable
private fun SavingsAccountsDialog(
    state: SavingsAccountState,
    onAction: (SavingsAccountAction) -> Unit,
) {
    when (state.dialogState) {
        is SavingsAccountState.DialogState.Error -> {
            AlertDialog(
                title = { Text("Error") },
                text = { Text(text = state.dialogState.message) },
                confirmButton = {
                    TextButton(
                        onClick = { onAction.invoke(SavingsAccountAction.CloseDialog) },
                    ) {
                        Text(stringResource(Res.string.feature_client_dialog_action_ok))
                    }
                },
                onDismissRequest = {},
            )
        }

        SavingsAccountState.DialogState.Loading -> MifosCircularProgress()

        else -> null
    }
}
```

TheKalpeshPawar (changes requested):

Arinyadav1:
@itsPronay 

Remove custom Note Empty Screen that you made as we already have this component. Use MifosEmptyCard,

please understand we about it what  you want to change.

TheKalpeshPawar (approved):

itsPronay (approved):

PR #575, opened 2017-03-10
style: Incorrect error message is displayed in the Centers activity

AbhilashG97:
Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Apply the `MifosStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #2490, opened 2025-09-02
Client Collateral Datas

ankitkumarrain:
Fixes - [Jira-#575](https://mifosforge.jira.com/browse/MIFOSAC-575)

This is my first pr.
Main Goal : It was to implement collateral data modules in the client profile . 

I have created a new directory in which i have CollateralUistate , CollateralViewModel, CollateralScreen under under the feature >loan module .


feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralScreen.kt
```diff
@@ -0,0 +1,152 @@
...
+            TopAppBar(title = {
+                val titleText = when (val state = uiState) {
+                    is ClientCollateralUiState.Success -> "Collateral Data (${state.totalItems} ${if (state.totalItems == 1) "Item" else "Items"})"
+                    is ClientCollateralUiState.Empty -> "Collateral Data (0 Items)"
+                    else -> "Collateral Data"
+                }
+                Text(text = titleText)
+            })
+        }
+    ) { paddingValues ->
```
TheKalpeshPawar: ## **Read my last three comments first. Then read this one and rest.**

---
You don't have to create your own scaffold, use MIfosScaffold.
Check the code of siblings screens and similar screens to see how we are using it.
If you want to use a component, first check if there is a Mifos component already created for it.
You will find all the created components in the ui module in core module.

You can check the sibling feature screens to see what component they are using. If you don't find one only then create a new component in the current screen's package.

TheKalpeshPawar (changes requested):

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralScreen.kt
```diff
@@ -0,0 +1,152 @@
...
+            Spacer(modifier = Modifier.height(4.dp))
+            Text("Total Collateral Value: ${item.totalCollateralValue}", style = MaterialTheme.typography.bodyMedium)
+            Box(modifier = Modifier.fillMaxWidth(), contentAlignment = Alignment.CenterEnd) {
+                 IconButton(onClick = onActionClick) {
+
+                }
+            }
+        }
+    }
+}
```
TheKalpeshPawar: Use MIfosActiosCollateralDataListingComponent composable here.
It's an already created composable in inside the package com.mifos.core.ui.components.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralScreen.kt
```diff
@@ -0,0 +1,152 @@
...
+        modifier = Modifier.padding(16.dp)
+    ) {
+        Text("Error: $message", color = MaterialTheme.colorScheme.error, style = MaterialTheme.typography.bodyLarge)
+        Spacer(modifier = Modifier.height(16.dp))
+        Button(onClick = onRetry) {
+            Text("Retry")
+        }
+    }
+}
+
```
TheKalpeshPawar: Use the MifosErrorComponent here like this

```kotlin
@Composable
private fun ClientProfileDialogs(
    state: ClientProfileState,
    onRetry: () -> Unit,
) {
    when (state.dialogState) {
        is ClientProfileState.DialogState.Loading -> MifosProgressIndicator()

        is ClientProfileState.DialogState.Error -> {
            MifosErrorComponent(
                isNetworkConnected = state.networkConnection,
                message = state.dialogState.message,
                isRetryEnabled = true,
                onRetry = {
                    onRetry()
                },
            )
        }

        null -> Unit
    }
}
```

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralScreen.kt
```diff
@@ -0,0 +1,152 @@
...
+            modifier = Modifier
+                .padding(32.dp)
+                .fillMaxWidth(),
+            horizontalAlignment = Alignment.CenterHorizontally,
+            verticalArrangement = Arrangement.Center
+        ) {
+            Text("No Item Found", style = MaterialTheme.typography.headlineSmall)
+        }
+    }
+}
```
TheKalpeshPawar: Use MifosEmptyCard instead.

like this
```kotlin
if (state.recurringDepositAccounts.isEmpty()) {
     MifosEmptyCard(msg = stringResource(Res.string.client_empty_card_message))
} else {
     LazyColumn {
```

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralViewmodel.kt
```diff
@@ -0,0 +1,78 @@
...
+                quantity = quantity,
+                unitValue = networkItem.basePrice,
+                totalCollateralValue = quantity * networkItem.basePrice
+            )
+        } else {
+
+            null
+        }
+    }
+}
```
TheKalpeshPawar: @ankitkumarrain 
search on internet about MVI architecture and learn about it.
See how and why we separate ui actions, events and ui state.
Then see inside the viewmodel of other screens, how we are doing it.

If you need help then just ask.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralViewmodel.kt
```diff
@@ -0,0 +1,78 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
+
+class ClientCollateralViewModel(
+    private val savedStateHandle: SavedStateHandle, // Assuming we might need clientId/groupId from nav args
+    private val clientDetailsRepository: ClientDetailsRepository
+) : ViewModel() {
+
```
TheKalpeshPawar: Use BaseViewModel class. It is present inside the com.mifos.core.ui.util package.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralScreen.kt
```diff
@@ -0,0 +1,152 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.dp
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+// Koin ViewModel import
+import org.koin.compose.viewmodel.koinViewModel
+
```
TheKalpeshPawar: You needed help with navigation right.

So, in this project we are using type safe navigation and also nested nav graphs.

I will give you a simple analogy.
Consider a Navgraph like your house (collection of multiple rooms). All houses have a door to enter inside, (now don't think what about windows or balcony), and it always opens inside one room always. Just like that a navgraph is a collection of multiple navigation destinations (screen). When you navigate to a navgraph there is always a screen set a startdestination that opens first..
And just like from inside of your room you can go inside multiple other room, so, just like that you can go to multiple other screens from that screen.

You won't create a navgraph here.

Learn about typesafe navigation and then see how we are using it.
In typesafe navigation you use serialized data classes instead of string route like in web or in normal string based navigation.
Here is an example:
```kotlin
@Serializable
data class ClientProfileRoute(
    val id: Int = -1
)
```
Instead of using a string you will use such data classes.  The arguments you pass along with string routs are instead passed a parameters to the data class.

Also learn about **extension functions**.

You will create a navigation destination(route) by creating a extension function on the **NavGraphBuilder**, something like this
_**NavGraphBuilder.navigateToCleintCollateralRoute**_

Just look into the client profile screen and see how navigation is done there.

If you need help ask.

TheKalpeshPawar (changes requested):

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralViewmodel.kt
ankitkumarrain: I think we have to implement mvvm pattern

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralViewmodel.kt
```diff
@@ -0,0 +1,127 @@
...
+                mutableStateFlow.update {
+                    it.copy(
+                        isLoading = false,
+                        accounts = result as List<CollateralItem>,
+                        dialogState = null
+
+
+                    )
+                }
+            }catch(e : Exception){
```
TheKalpeshPawar: Ankit ```repository.getCollateralItems(route.clientId)``` returns ```DataState<List<CollateralItem>>```, and you are casting it as ```List<CollateriaItem>```, that is not a valid operation.

See in other viewmodels how we are handling DataState and how to retrieve data from it.

This is the reason you are not getting any data.

TheKalpeshPawar (changes requested):

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralViewmodel.kt
```diff
@@ -0,0 +1,127 @@
...
+        when (action) {
+            is collateralAction.cardClicked -> handleCardClicked(action.activeIndex)
+            collateralAction.toggleFiler -> toggleFiler()
+            collateralAction.toggleSearchBar -> toggleSearchBar()
+            is collateralAction.viewAccount -> sendEvent(collateralEvent.viewAccount(action.accountId))
+            collateralAction.refresh -> fetchAllCollateralAccount()
+
+        }
+
+    }
```
TheKalpeshPawar: Always write class, interfaces , composable function and objects name in `Pascal Case` Eg: `CollateralAction.ToggleFilter`

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralViewmodel.kt
```diff
@@ -0,0 +1,127 @@
...
+        fetchAllCollateralAccount()
+    }
+    private fun fetchAllCollateralAccount(){
+        viewModelScope.launch {
+            mutableStateFlow.update{
+                it.copy(
+                    isLoading = true,
+                )
+
+            }
```
TheKalpeshPawar: Do not use a separate state parameter for handling loading. See how we are using DialogState for handling Error and Loading on UI.
You will find something like this in other UiState classes.
```kotlin
data class SomeUiState(
    val dialogState: DialogState? = null
 ){
     sealed interface DialogState {
         object Loading: DialogState
         data class Error(val message: String): DialogState
    }
}
```

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralViewmodel.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    }
+    private fun toggleSearchBar() {
+        mutableStateFlow.update {
+            it.copy(
+                isSearchBarActive = !state.isSearchBarActive,
+            )
+        }
+
+    }
```
TheKalpeshPawar: Write !it.isFilterAcitve and !it.isSearchBarActive instead of using state.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/ClientCollateral/ClientCollateralViewmodel.kt
```diff
@@ -0,0 +1,127 @@
...
+        mutableStateFlow.update {
+            it.copy(
+                isCardActive = !state.isCardActive,
+                currentlyActiveIndex = index,
+            )
+        }
+
+    }
+
+
```
TheKalpeshPawar: Same here.

TheKalpeshPawar (changes requested):

amanna13:
@TheKalpeshPawar Is this issue fixed ?? The [Jira-#575](https://mifosforge.jira.com/browse/MIFOSAC-575) ticket is marked as Done and also currently we have this feature in the android-client. Though we dont have certain things like - 
``` Collateral Data Actions:
Available actions should include:
View Account → Navigates to detailed collateral data screen. 
``` 
as mentioned in the ticket

PR #2491, opened 2025-09-03
feat:New Loan Account – Details Step Page

revanthkumarJ:
Fixes - [Jira-#528](https://mifosforge.jira.com/browse/MIFOSAC-528)

https://github.com/user-attachments/assets/73f367dd-60d5-4bec-8c61-014c39a40127

### DIFF
--- a/feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountScreen.kt
+++ b/feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountScreen.kt
@@ -59,7 +59,7 @@ internal fun NewLoanAccountScreen(
 
     NewLoanAccountDialogs(
         state = state,
-        onRetry = { viewModel.trySendAction(NewLoanAccountAction.NextStep) },
+        onRetry = { viewModel.trySendAction(NewLoanAccountAction.Retry) },
     )
 
     NewLoanAccountScaffold(
--- a/feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountScreen.kt
+++ b/feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountScreen.kt
@@ -132,6 +132,16 @@ private fun NewLoanAccountScaffold(
                     )
                 }
             }
+
+            NewLoanAccountState.ScreenState.NetworkError -> {
+                MifosErrorComponent(
+                    isNetworkConnected = state.networkConnection,
+                    isRetryEnabled = true,
+                    onRetry = {
+                        onAction(NewLoanAccountAction.Retry)
+                    },
+                )
+            }
         }
         if (state.isOverLayLoadingActive) {
             MifosProgressIndicatorOverlay()
@@ -147,7 +157,6 @@ private fun NewLoanAccountDialogs(
     when (state.dialogState) {
         is NewLoanAccountState.DialogState.Error -> {
             MifosErrorComponent(
-                isNetworkConnected = state.networkConnection,
                 message = state.dialogState.message,
                 isRetryEnabled = true,
                 onRetry = {
--- a/feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountScreen.kt
+++ b/feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountScreen.kt
@@ -133,6 +133,9 @@ private fun NewLoanAccountScaffold(
                 }
             }
         }
+        if (state.isOverLayLoadingActive) {
+            MifosProgressIndicatorOverlay()
+        }
     }
 }
 
@@ -152,8 +155,6 @@ private fun NewLoanAccountDialogs(
                 },
             )
         }
-
-        NewLoanAccountState.DialogState.LoadingOverLay -> MifosProgressIndicatorOverlay()
 
         null -> Unit
     }
--- a/feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountScreen.kt
+++ b/feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountScreen.kt
@@ -29,6 +29,7 @@ import com.mifos.core.designsystem.component.MifosScaffold
 import com.mifos.core.ui.components.MifosBreadcrumbNavBar
 import com.mifos.core.ui.components.MifosErrorComponent
 import com.mifos.core.ui.components.MifosProgressIndicator
+import com.mifos.core.ui.components.MifosProgressIndicatorOverlay
 import com.mifos.core.ui.components.MifosStepper
 import com.mifos.core.ui.components.Step
 import com.mifos.core.ui.util.EventsEffect
@@ -151,6 +152,8 @@ private fun NewLoanAccountDialogs(
                 },
             )
         }
+
+        NewLoanAccountState.DialogState.LoadingOverLay -> MifosProgressIndicatorOverlay()
 
         null -> Unit
     }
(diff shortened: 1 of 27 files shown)


feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountScreen.kt
```diff
@@ -0,0 +1,156 @@
...
+            }
+        },
+    )
+
+    MifosScaffold(
+        title = stringResource(Res.string.new_loan_account_title),
+        onBackPressed = { onAction(NewLoanAccountAction.NavigateBack) },
+        modifier = modifier,
+    ) { paddingValues ->
+        if (state.dialogState == null) {
```
Nagarjuna0033: @revanthkumarJ This should not be like this, dialogs should not interrupt the screen.

In this case if dialog is shown the screen will be empty

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountScreen.kt
```diff
@@ -0,0 +1,156 @@
...
+    }
+}
+
+@Composable
+private fun NewLoanAccountDialogs(
+    state: NewLoanAccountState,
+    onRetry: () -> Unit,
+) {
+    when (state.dialogState) {
+        is NewLoanAccountState.DialogState.Loading -> MifosProgressIndicator()
```
Nagarjuna0033: Loading should not be in dialog

Nagarjuna0033 (changes requested):

Nagarjuna0033:
@revanthkumarJ See These changes going to break, see you are observing the network, and at the same time you are calling the other function in the `init block` concurrently so what is user doesn't have any internet connection still this will hit the Api and throws error which not handled properly in this case. and also validation is not up to mark. there is a field `externalId` you are just changing the value by invoking action is emojis and other invalid characters are allowed? if not you need to validate all those which is not handled in this pr and in upcoming prs also and why don't you separate the validations from `onHandle` function you are just implementing the logic in the same block which will be not readable. for example you ae doing like this 

```
is NewLoanAccountAction.OnExternalIdChange -> {
                mutableStateFlow.update {
                    it.copy(
                        externalId = action.value,
                    )
                }
            }
```
            
but you can do like this 
            
```
is NewLoanAccountAction.OnExternalIdChange -> handleExternalIdChange()
```
            
            
in the function you can validate every edge case and udpate the errors if any.

validations should be correctly handled and try out the feature in all cases like if user doesn't have internet or any other exception
            
``` write previews for all states```

File: feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountViewModel.kt [hunk 3]
```diff
@@ -0,0 +1,268 @@
...
+
+internal class NewLoanAccountViewModel(
+    private val getAllLoanUseCase: GetAllLoanUseCase,
+    private val getLoansAccountTemplateUseCase: GetLoansAccountTemplateUseCase,
+    private val networkMonitor: NetworkMonitor,
+    val savedStateHandle: SavedStateHandle,
+) :
+    BaseViewModel<NewLoanAccountState, NewLoanAccountEvent, NewLoanAccountAction>(
+        initialState = run {
+            NewLoanAccountState(clientId = savedStateHandle.toRoute<NewLoanAccountRoute>().clientId)
```
Nagarjuna0033, line 37: @revanthkumarJ don't we have client id in the datastore

Nagarjuna0033 (changes requested):

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountViewModel.kt
revanthkumarJ: no as in android client field officer can see all clients client id is passed based on which client he clicked

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountViewModel.kt
```diff
@@ -0,0 +1,313 @@
...
+    val submissionDate: String = DateHelper.getDateAsStringFromLong(Clock.System.now().toEpochMilliseconds()),
+    val showSubmissionDatePick: Boolean = false,
+    val expectedDisbursementDate: String = DateHelper.getDateAsStringFromLong(Clock.System.now().toEpochMilliseconds()),
+    val showExpectedDisbursementDatePick: Boolean = false,
+    val linkSavingsIndex: Int = -1,
+    val isCheckedStandingInstructions: Boolean = false,
+) {
+    sealed interface DialogState {
+        data class Error(val message: String) : DialogState
+        data object LoadingOverLay : DialogState
```
Nagarjuna0033: @revanthkumarJ I think this also should not be in dialog. can you confirm it once

revanthkumarJ: updated

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountViewModel.kt
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    private fun handleProductNameChange(action: NewLoanAccountAction.OnProductNameChange) {
+        mutableStateFlow.update { it.copy(loanProductSelected = action.index) }
+        loadLoanAccountTemplate(state.productLoans[action.index].id ?: -1)
+    }
+
+    private fun handleExternalIdChange(action: NewLoanAccountAction.OnExternalIdChange) {
+        mutableStateFlow.update { it.copy(externalId = action.value) }
+    }
```
Nagarjuna0033, line 130: @revanthkumarJ As I said you should validate it, user may enter emojis or other invalid characters

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountViewModel.kt
```diff
@@ -0,0 +1,317 @@
...
+                mutableStateFlow.update {
+                    it.copy(networkConnection = isConnected)
+                }
+                if (isConnected) {
+                    loadAllLoans()
+                } else {
+                    mutableStateFlow.update {
+                        it.copy(
+                            dialogState = NewLoanAccountState.DialogState.Error(""),
+                        )
```
Nagarjuna0033, line 193: Why empty message show something

Nagarjuna0033 (changes requested):

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountViewModel.kt
revanthkumarJ: it automatically shows "No internet" as we are passing internet status here

revanthkumarJ: as it is invoked when no internet and MifosErrorComponent ignores the msg and shows "No Internet Text" irrespective of what error msg we send so i send empty msg

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountViewModel.kt
revanthkumarJ: i have validated on on submit

revanthkumarJ: as there are more than 50 text fields here ( in this screen ) it is not needed to check everytime a text field is changed

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/newLoanAccount/NewLoanAccountViewModel.kt
Nagarjuna0033: Understand but this can also be moved to screen state introduce a variable `Network` and render if there is no network

Nagarjuna0033 (approved):

Issue #525, opened 2017-02-16
UX: `Connection Settings` on login screen doesn't feel clickable.


devkhan:
**Summary:** 

The `Connection Settings` on the login screen allows to change connection option, but it feels like a plain `TextView` when it should be a `Button` or `ImageButton`.

**Steps to reproduce:** 

NA

**Expected behavior:** 

NA

**Observed behavior:** 

NA

**Device and Android version:** 

Device independent.

**Screenshots:** 

NA

singhalok641:
I am fixing this issue

singhalok641:
I am fixing this issue

Pritom14:
Please review #546 . 
Here I have changed the TextView into a Button

Pritom14:
I am new to this and whenever I edit one of the files another gets created with ~ at the end. I have submitted another pr #547 , but it has the same problem it has two extra files. Can you help me fix it?

Pritom14:
Please review #548

therajanmaurya:
fix in #552

PR #2497, opened 2025-09-06
feat : New Loan Account – Charges Step Implementation

revanthkumarJ:
Fixes - [Jira-#525](https://mifosforge.jira.com/browse/MIFOSAC-525)

https://github.com/user-attachments/assets/95bf1455-59ea-4afe-8d62-516dd883437f


itsPronay (approved):
i’m not sure if making a separate function for every single action is a good idea, since it makes the viewmodel look messy...
Other than that everything looks good to me

PR #2501, opened 2025-09-10
MIFOSAC-577 Client Profile Documents

TheKalpeshPawar:
- **Client Profile Documents - [Jira-#MIFOSAC-577](https://mifosforge.jira.com/browse/MIFOSAC-577)**
---

##Video
<video src="https://github.com/user-attachments/assets/b652ac03-c985-458c-9720-2ce31941e8b4"/>

---
<p><s>
## Goal not achieved yet: <br>
**Show preview for pdf documents.**<br>
I haven't found a kmp library yet for this. <br>
Actually I found one this morning, but it doesn't suppot IOS <br>
Link: [compose-pdf](https://klibs.io/project/zt64/compose-pdf) <br>
But I was busy making the screen work, I haven't properly looked into this issue yet. <br>
 </s></p>
---

## Important things to note
- The maximum upload size for a document is restricted to **1MB**.
- The user are restricted to  **JPG** ,  **JPEG**, **PNG** AND **PDF** format files for uploading.
- Documents (both image and PDFs) download by user to view are stored in the app cache. Also the PDF files which the user chooses to preview by clicking on **View** button are also stored in the cache. But this doesn't fill the cache much because each saved file overwrites the existing file. Atmost there will be only two file in the app cache **attachment.pdf** and **attachment.jpeg**.

### Image as Documents
- Image files preview is shown in the app itself.
- It is shown when uploading, when user wished to preview again once before when uploading,
in which case they can update the document too.
- When the user download the Image file from the server it is stored in the app cache and the preview is shown in the app itself.

### PDF as Documents
- PDF file preview is not shown when the user picks them first time for upload,
- The PDF file will open in an external app if the user wished to preview it once (by clicking on View button)
before upload. They can replace the file too.
- When the user downloads the PDF file from the server it is stored in the app cache and the a chooser for opening the file in external PDF viewer apps is shown.
---

## What I have done in this PR ?
My task was to create feature for clients documents upload and preview.

- This feature is divided into three screens
   -- ClientDocumentList Screen: This screen lists documents of the client.
   -- ClientAddDocument Screen: This screen is for adding and updating client documents.
   -- DocumentPreviewScreen: This is for show preview of document and choosing a new one if needed.

One thing about the DocumentPreview Screen, it is used with other screen too, therefore I wanted it to be a separate screen.
I don't know If I have succeeded in doing so.

Since, these are three different screens they have their own different states and also their own view model.
As they were state dependent on each other, I had to find out some way of sharing states between these three screens.

Navigation arguments were a terrible option.
I tried using shared viewmodel then, which also didn't work, because of too many states and having to host a shared view model in a navgraph, which created issue with navigation.

---
## My solution
**I have used a repository and hosted common states in it and injected it in these three screens.**
---


Arinyadav1:
Make a toggle animation only for one card at a time using Id, if one card animation show then we click on another card then previous one must to close automatically. 

take a reference from note list screen in feature module.

Arinyadav1:
Document list must have automatically update when we navigate from add/update screen don't use manually refresh approach.

sam-arth07:
Also the list could be sorted in desc order. So the latest uploaded document will be at the top and easily accessible

TheKalpeshPawar:
> Also the list could be sorted in desc order. So the latest uploaded document will be at the top and easily accessible

Done.

TheKalpeshPawar:
> Document list must have automatically update when we navigate from add/update screen don't use manually refresh approach.

Done

TheKalpeshPawar:
> Make a toggle animation only for one card at a time using Id, if one card animation show then we click on another card then previous one must to close automatically.
> 
> take a reference from note list screen in feature module.

done

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientAddDocuments/ClientAddDocumentScreenViewmodel.kt
```diff
@@ -0,0 +1,352 @@
...
+        }
+    }
+
+    private fun pickFromGallery() {
+        viewModelScope.launch {
+            documentSelectAndUploadRepository.selectImageFromGallery()
+                .collect { dataState ->
+                    when (dataState) {
+                        is DataState.Error<*> -> {
+                            mutableStateFlow.update {
```
revanthkumarJ, line 115: remove <*> if you find any other in other places too

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientAddDocuments/ClientAddDocumentScreenViewmodel.kt
```diff
@@ -0,0 +1,352 @@
...
+        }
+    }
+
+    private fun pickFromFiles() {
+        viewModelScope.launch {
+            documentSelectAndUploadRepository.selectImageFromFile()
+                .collect { dataState ->
+                    when (dataState) {
+                        is DataState.Error<*> -> {
+                            mutableStateFlow.update {
```
revanthkumarJ, line 161: here too <*>

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientAddDocuments/ClientAddDocumentScreenViewmodel.kt
```diff
@@ -0,0 +1,352 @@
...
+                    ).collect { dataState ->
+                        when (dataState) {
+                            is DataState.Error<*> -> {
+                                errorDialogState(dataState.message)
+                            }
+                            DataState.Loading -> {
+                                loadingDialogState()
+                            }
+                            is DataState.Success<*> -> {
+                                nullDialogState()
```
revanthkumarJ, line 222: here also <*>

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientAddDocuments/ClientAddDocumentScreenViewmodel.kt
```diff
@@ -0,0 +1,352 @@
...
+
+    private fun previewPdfInExternalApp(platformFile: PlatformFile) {
+        viewModelScope.launch {
+            loadingDialogState()
+            try {
+                openPdfWithDefaultExternalApp(platformFile)
+                nullDialogState()
+            } catch (e: Exception) {
+                errorDialogState(e.message ?: "Unknown error")
+            }
```
revanthkumarJ, line 292: move Unknown error also to string resources

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientAddDocuments/ClientAddDocumentsScreen.kt
```diff
@@ -0,0 +1,360 @@
...
+            text = if (state.step == EntityDocumentState.Step.ADD) {
+                stringResource(Res.string.no_file_selected)
+            } else {
+                state.pickedDocumentName
+            },
+            style = MaterialTheme.typography.bodyLarge,
+            fontFamily = FontFamily.SansSerif,
+            modifier = Modifier.padding(
+                start = 16.dp,
+                top = 18.dp,
```
revanthkumarJ: use DesignToken.padding. those

revanthkumarJ: in figma if it had something like 18 or 19 or 23  don't use it.

use standard like 2,4,8,12,16 only

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientAddDocuments/ClientAddDocumentsScreen.kt
```diff
@@ -0,0 +1,360 @@
...
+            ),
+            modifier = Modifier
+                .padding(end = 16.dp)
+                .height(36.dp)
+                .wrapContentWidth(),
+        ) {
+            Text(
+                text = if (state.step == EntityDocumentState.Step.ADD) {
+                    "Add"
+                } else {
```
revanthkumarJ, line 354: move to string resources

revanthkumarJ (changes requested):

revanthkumarJ (approved):

PR #2519, opened 2025-10-06
Profile photo not refreshing after upload in Client Profile

ufuomaisaac:
Fixes - [Jira-#Issue_Number](https://mifosforge.jira.com/browse/MIFOSAC-595)

## Summary by CodeRabbit

* **New Features**
  * Introduced enhanced status dialogs with visual indicators (success with checkmark icon and green color, failure with error icon and red color) for clearer user feedback during operations.
  * Implemented automatic client data synchronization to refresh profile information when updates occur.


ufuomaisaac:
https://github.com/user-attachments/assets/f7a75d46-b64a-465a-8198-c77061b89fcb

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientEditProfile/ClientProfileEditScreen.kt
```diff
@@ -76,14 +76,14 @@ internal fun ClientProfileEditScreen(
 ) {
     val state by viewModel.stateFlow.collectAsStateWithLifecycle()
 
-    EventsEffect(viewModel.eventFlow) { event ->
+    /*EventsEffect(viewModel.eventFlow) { event ->
         when (event) {
             ClientProfileEditEvent.NavigateBack -> onNavigateBack()
             ClientProfileEditEvent.OnSaveSuccess -> {
                 onNavigateBack()
```
revanthkumarJ: why commented this ?

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientEditProfile/ClientProfileEditScreen.kt
```diff
@@ -206,6 +206,16 @@ private fun ClientProfileEditDialogs(
             )
         }
 
+        is ClientProfileEditState.DialogState.Success -> {
+            MifosBasicDialog(
+                title = "Success",
+                content = { Text("Your profile image has been updated successfully!") },
+                confirmText = "Done",
```
revanthkumarJ: use Success and Failure Dilaogs instead of Basic Dialog.

revanthkumarJ (changes requested):

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientEditProfile/ClientProfileEditScreen.kt
revanthkumarJ: This one

MifosStatusDialog

which accepts success and Failure state and also move strings to string resources

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientEditProfile/ClientProfileEditScreen.kt
ufuomaisaac: I commented it because it is not needed for the navigation, and it is excess code
The screen navigation works well with or without it

revanthkumarJ: Remove it if not needed

cmp-android/src/main/res/values/strings.xml
```diff
@@ -421,6 +421,10 @@
     <string name="no_more_clients_available">No more Clients Available</string>
     <string name="feature_groups_no_more_groups_available">No more Groups Available</string>
     <string name="no_more_centers_available">No more Centers Available</string>
+    <string name="feature_client_profile_update_success_message">Your profile image has been updated successfully!</string>
+    <string name="feature_client_profile_update_error_message">We couldn’t update your profile image. Please try again.</string>
+    <string name="should_refresh">should_refresh</string>
```
biplab1: This should be added to feature:client's commonMain strings.xml (`android-client\feature\client\src\commonMain\composeResources\values\strings.xml`)

ufuomaisaac: Oh okay, I will do that and push it

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientEditProfile/ClientProfileEditScreen.kt
ufuomaisaac: I am done with the changes

cmp-android/src/main/res/values/strings.xml
ufuomaisaac: I am apply the changes

core/ui/src/commonMain/kotlin/com/mifos/core/ui/util/Constants.kt
```diff
@@ -0,0 +1,3 @@
+package com.mifos.core.ui.util
+
+internal const val PROFILE_SHOULD_REFRESH_KEY = "profile_should_refresh"
```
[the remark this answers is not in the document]
ufuomaisaac: Done

feature/client/src/commonMain/kotlin/com/mifos/feature/client/utils/Constants.kt
```diff
@@ -0,0 +1,3 @@
+package com.mifos.feature.client.utils
+
+internal const val PROFILE_SHOULD_REFRESH_KEY = "profile_should_refresh"
```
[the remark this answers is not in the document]
ufuomaisaac: Done

core/ui/src/commonMain/kotlin/com/mifos/core/ui/util/Constants.kt
```diff
@@ -0,0 +1,3 @@
+package com.mifos.core.ui.util
+
+const val PROFILE_SHOULD_REFRESH_KEY = "profile_should_refresh"
```
itsPronay: Creating a whole file just for one const is unnecessary. Check `core/common`, we already have Constants there; if we need it, put it in there and use it.

ufuomaisaac:
Okay, I will do that
Thank you Pronay

kartikey004:
@ufuomaisaac, the navigation flag approach is fine but i think it depends on the navigation backstack. I thought of using a repository level broadcast instead so the profile screen automatically refreshes whenever the client is updated without relying on navigation data. this should help us keep the screens independent and makes it easier to handle updates if more screens change the client in the future. also it wont require much changes to multiple files.

ufuomaisaac:
A repository level board cast? 
I am not even sure what that means, does it is means, I should done it in the repository instead on the view model and the screen, I don't think that is really needed for a navigation, I would prefer if you could explain how you would have done it better

Also I change a few things, which is the constants, added a new dialog because I was given a new dialog to display

I would indulge you to explain your approach and what you mean by Repository level board cast, also I would appreciate if you discuss your approach on the group so that people can also contribute to the approach we are to use, moving forward

kartikey004:
I mean exposing a signal from the `ClientDetailRepository.kt` that emits whenever the client data is updated. any screen that displays client data can simply listen to this signal and refresh when it receives an update.

This way the UI does not need to know where the update came from or worry about passing flags back through the backstack.

ufuomaisaac:
@kartikey004 The issue is not that the data in the UI doesn't get updated/ the UI doesn't get updated through the view model; it does. The main issue here is that when the user goes back to the previous screen  through the back button, the previous screen doesn't get updated with the current data from the view model

The solution to solve this problem is to refresh the previous screen so that it can load the updated data from the view model

There are different ways to solve this problem, but I chose to solve it by passing flags through the back stack to check if the client image was updated or not, then it refreshes the previous screen based on that answer

Another optimized way to always load the previous screen with the updated  view model data, when we navigate to the previous screen, the screen loads the data from the view model instead of just displaying the outdated data from the view model

ufuomaisaac:
@kartikey004  So I actually do not think a repository that emits signals is necessary in this instance. We just need to refresh the previous screen as we go back through the stacks, so that the screen displays the updated data from the view model

let me know what you think...

kartikey004:
> @kartikey004 So I actually do not think a repository that emits signals is necessary in this instance. We just need to refresh the previous screen as we go back through the stacks, so that the screen displays the updated data from the view model
> 
> let me know what you think...

@ufuomaisaac, I agree with you that the issue is refreshing the previous screen so it shows the latest data from the view model.My concern was mainly that the flag approach works now, but it adds some manual handling that can be easy to miss if the flow changes later. 

repository signal will automatically trigger a refresh whenever the client profile changes without depending on navigation, which feels simpler to maintain. also as the bot pointed out, the current PR has a few edge cases and coupling issues around the backstack. The flag approach is fine and works but difficult to maintain. thats what i meant.

Nagarjuna0033:
@ufuomaisaac Hi, I think you don't need any flag usage instead of manual resetting and setting flags, go for repository level signal as code rabbit suggested as it will be clean and better approach and also we can reuse it if we need it in any other places.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientProfile/ClientProfileScreen.kt
```diff
@@ -62,6 +65,20 @@ internal fun ClientProfileScreen(
...
+        ?.getStateFlow(PROFILE_SHOULD_REFRESH_KEY, false)
+        ?.collectAsStateWithLifecycle(initialValue = false)
+        ?: remember { mutableStateOf(false) }
+
+    LaunchedEffect(profileUpdated) {
+        if (profileUpdated) {
+            viewModel.trySendAction(ClientProfileAction.OnRetry)
+            savedStateHandle?.set(PROFILE_SHOULD_REFRESH_KEY, false) // reset after refresh
+        }
+    }
```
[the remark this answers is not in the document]
ufuomaisaac: yes

ufuomaisaac:
@Nagarjuna0033 @kartikey004 Alright, will work on repository-level signal, thanks for your input

kartikey004:
@ufuomaisaac, I think you should remove the leftover changes from the earlier flag based approach to keep the PR clean for merging. also it would be good to apply the repository signal to all screens that display the profile picture so they stay in sync.

also please read the contribution guide once, it will help you a lot.

> In MacOS, Windows or Linux, you should run the following commands before opening a PR, and make sure to pass all the commands:

`./gradlew check -p build-logic` this checks build-logic configured properly.)
`./gradlew spotlessApply --no-configuration-cache` an check and apply formatting to any file.
`./gradlew dependencyGuardBaseline` to generate dependency-guard baseline.
`./gradlew detekt` to check detekt error.
`./gradlew build` to build the project.
`./gradlew testDebug :lint:test :androidApp:lintRelease :lint:lint` to check lint and test error.)
`./gradlew updateReleaseBadging` to update the badging for the project.)

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientProfile/ClientProfileScreen.kt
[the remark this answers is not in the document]
ufuomaisaac: Also solve for ClientProfileEditScreen and ClientProfileDetailsScreen

core/ui/src/commonMain/kotlin/com/mifos/core/ui/util/Constants.kt
ufuomaisaac: I have removed the file

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/ClientDetailsEditRepositoryImpl.kt
```diff
@@ -12,10 +12,23 @@ package com.mifos.core.data.repositoryImp
...
+        replay = 0,
+        extraBufferCapacity = 1
+    )
+
+    override val clientDataUpdated: SharedFlow<Int> = _clientDataUpdated.asSharedFlow()
+
+
     override suspend fun updateClient(clientId: Int, clientPayload: ClientPayloadEntity): Int? {
         return dataManagerClient.updateClient(clientId, clientPayload)
     }
```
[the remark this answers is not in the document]
ufuomaisaac: Resolved!

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientDetailsProfile/ClientProfileDetailsScreen.kt
```diff
@@ -44,7 +44,9 @@ import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
```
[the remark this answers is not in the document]
ufuomaisaac: Resolved!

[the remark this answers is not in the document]
ufuomaisaac: Resolved!

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientProfile/ClientProfileScreen.kt
```diff
@@ -63,6 +65,7 @@ internal fun ClientProfileScreen(
 ) {
     val state by viewModel.stateFlow.collectAsStateWithLifecycle()
 
+
```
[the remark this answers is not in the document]
ufuomaisaac: yes

Issue #595, opened 2017-03-18
Create Client option inside Create Client


aj019:
**Summary:** 

Inside the Create Client Screen if we click on the toolbar menu. It shows Create client as one of the options. And when the user clicks on the Create Client option nothing happens.

**Steps to reproduce:** 

1) Open the app
2) From the navbar select clients
3) From the toolbar menu select Create Client
4) In Create client if you click on the toolbar menu you will get an option saying Create Client

**Expected behavior:** 

The Create client option inside Create client should either be invisible or should be have distinct colour depicting it to be disabled.

**Observed behavior:** 

When you click on the the Create Client option inside Create Client nothing happens.

**Device and Android version:** 

Nexus 5 API 23

**Screenshots:** 
<img width="247" alt="screen shot 2017-03-18 at 7 18 55 am" src="https://cloud.githubusercontent.com/assets/7893859/24068083/5a98d460-0bac-11e7-89cc-92e04ecdce85.png">

therajanmaurya:
duplicate https://github.com/openMF/android-client/issues/526

PR #2570, opened 2026-01-05
fix: Client profile photo and details not refreshing after update

kartikey004:
**Fixes** - [Jira-#595](https://mifosforge.jira.com/browse/MIFOSAC-595)

**Description** - The client profile was not updating when navigating back from the edit screen. I implemented a repository-level broadcast flow so the profile screen listens for update signals and refreshes automatically.

Also added a status dialog box after photo uploads to confirm the changes.

**Video** - 

https://github.com/user-attachments/assets/760578b8-2f69-4f0d-bd1b-59a1f4395cf6

## Summary by CodeRabbit

* **New Features**
  * Client profiles now auto-refresh via a global client update event; edits and photo changes broadcast updates.
  * Added an explicit trigger to emit client update events for reactive updates.

* **UI/UX Improvements**
  * Redesigned status dialog with centered icon, optional action button, and auto-confirm on success.
  * Added localized success/failure titles and messages for profile photo operations and persistent dialog feedback.

### DIFF
--- a/core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosStatusDialog.kt
+++ b/core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosStatusDialog.kt
@@ -23,7 +23,6 @@ import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.wrapContentHeight
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.text.style.TextAlign
@@ -31,7 +30,6 @@ import com.mifos.core.designsystem.component.MifosButton
 import com.mifos.core.designsystem.theme.DesignToken
 import com.mifos.core.designsystem.theme.MifosTheme
 import com.mifos.core.designsystem.theme.MifosTypography
-import kotlinx.coroutines.delay
 import org.jetbrains.compose.resources.painterResource
 import org.jetbrains.compose.ui.tooling.preview.Preview
 
--- a/core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosStatusDialog.kt
+++ b/core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosStatusDialog.kt
@@ -26,7 +26,6 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.text.style.TextAlign
 import com.mifos.core.designsystem.component.MifosButton
 import com.mifos.core.designsystem.theme.DesignToken
--- a/core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosStatusDialog.kt
+++ b/core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosStatusDialog.kt
@@ -44,6 +44,7 @@ fun MifosStatusDialog(
     successMessage: String,
     failureTitle: String,
     failureMessage: String,
+    showButton: Boolean = true,
     modifier: Modifier = Modifier,
 ) {
     val (title, message, icon) = when (status) {
@@ -57,13 +58,6 @@ fun MifosStatusDialog(
             failureMessage,
             painterResource(Res.drawable.ic_icon_error),
         )
-    }
-
-    if (status == ResultStatus.SUCCESS) {
-        LaunchedEffect(Unit) {
-            delay(1500)
-            onConfirm()
-        }
     }
 
     Column(
@@ -95,7 +89,7 @@ fun MifosStatusDialog(
             )
         }
 
-        if (status == ResultStatus.FAILURE) {
+        if (showButton) {
             MifosButton(
                 onClick = onConfirm,
                 text = {
--- a/core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosStatusDialog.kt
+++ b/core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosStatusDialog.kt
@@ -15,19 +15,24 @@ import androidclient.core.ui.generated.resources.ic_icon_success
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.wrapContentHeight
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.text.style.TextAlign
 import com.mifos.core.designsystem.component.MifosButton
 import com.mifos.core.designsystem.theme.DesignToken
 import com.mifos.core.designsystem.theme.MifosTheme
 import com.mifos.core.designsystem.theme.MifosTypography
+import kotlinx.coroutines.delay
 import org.jetbrains.compose.resources.painterResource
 import org.jetbrains.compose.ui.tooling.preview.Preview
 
@@ -53,6 +58,13 @@ fun MifosStatusDialog(
             failureMessage,
             painterResource(Res.drawable.ic_icon_error),
         )
+    }
+
+    if (status == ResultStatus.SUCCESS) {
+        LaunchedEffect(Unit) {
+            delay(1500)
+            onConfirm()
+        }
     }
 
     Column(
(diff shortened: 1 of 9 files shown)


feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientEditProfile/ClientProfileEditScreen.kt
```diff
@@ -260,5 +266,31 @@ private fun ClientProfileEditDialogs(
                 selectedImageFileCallback = {},
             )
         }
+
+        is ClientProfileEditState.DialogState.ShowStatusDialog -> {
+            androidx.compose.ui.window.Dialog(
```
itsPronay: fix import

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientEditProfile/ClientProfileEditViewModel.kt
```diff
@@ -138,7 +155,17 @@ internal class ClientProfileEditViewModel(
...
                         )
                     }
+                    clientDetailsRepo.triggerClientUpdate()
                     loadImage(route.id)
+                    mutableStateFlow.update {
+                        it.copy(
+                            openImagePicker = false,
+                            dialogState = ClientProfileEditState.DialogState.ShowStatusDialog(
+                                status = ResultStatus.SUCCESS,
+                                msg = "Profile photo updated successfully",
```
itsPronay: put in strings

itsPronay (changes requested):

ufuomaisaac:
@kartikey004 I noticed that you didn't navigate through the MifosBreadCrumb; you navigated through the back
 arrow at the top. We are to solve the problem by navigating backward through the breadcrumb, as I was instructed by Sir Rajan, not otherwise. Also, I just confirmed by checking out the development branch, the back arrow has been removed

kartikey004:
> @kartikey004 I noticed that you didn't navigate through the MifosBreadCrumb; you navigated through the back arrow at the top. We are to solve the problem by navigating backward through the breadcrumb, as I was instructed by Sir Rajan, not otherwise. Also, I just confirmed by checking out the development branch, the back arrow has been removed

@ufuomaisaac I used the back arrow only for the demo but the repository-level broadcast does not depend on navigation. since the Profile Screen listens directly to the data layer, it refreshes automatically no matter how the user goes back whether it’s the breadcrumb, the hardware back button or any other way.

ufuomaisaac:
Fairs

biplab1:
@kartikey004 Can you tell me if we really need a continue button in the success dialog? We don't have any other options for the user to interact with. We could simply load the screen details page after the dialog is briefly displayed. Let me know your thoughts on this.

image

kartikey004:
> @kartikey004 Can you tell me if we really need a continue button in the success dialog? We don't have any other options for the user to interact with. We could simply load the screen details page after the dialog is briefly displayed. Let me know your thoughts on this.
> 
@biplab1 I understand your point. this can reduce an extra step. but auto closing a dialog can feel a bit abrupt since it appears and disappears quickly and some users may miss the message. also different users read at different speeds so its hard to decide an optimal auto-close time for the dialog.

biplab1:
> > @kartikey004 Can you tell me if we really need a continue button in the success dialog? We don't have any other options for the user to interact with. We could simply load the screen details page after the dialog is briefly displayed. Let me know your thoughts on this.
> 
> @biplab1 I understand your point. this can reduce an extra step. but auto closing a dialog can feel a bit abrupt since it appears and disappears quickly and some users may miss the message. also different users read at different speeds so its hard to decide an optimal auto-close time for the dialog.

@kartikey004 A continue button makes sense for error dialogs, where users may need time to read and acknowledge the message.

For success messages, since we already show a success icon along with `Success` text indicating the upload completed successfully, a brief visibility should be sufficient for users to understand what happened, and we can navigate forward automatically.

kartikey004:
> > > @kartikey004 Can you tell me if we really need a continue button in the success dialog? We don't have any other options for the user to interact with. We could simply load the screen details page after the dialog is briefly displayed. Let me know your thoughts on this.
> > 
> > 
> > @biplab1 I understand your point. this can reduce an extra step. but auto closing a dialog can feel a bit abrupt since it appears and disappears quickly and some users may miss the message. also different users read at different speeds so its hard to decide an optimal auto-close time for the dialog.
> 
> @kartikey004 A continue button makes sense for error dialogs, where users may need time to read and acknowledge the message.
> 
> For success messages, since we already show a success icon along with `Success` text indicating the upload completed successfully, a brief visibility should be sufficient for users to understand what happened, and we can navigate forward automatically.

That makes sense. I'll remove the button and add a short delay before it navigates.

kartikey004:
@biplab1 is this flow correct for success dialog?

https://github.com/user-attachments/assets/dec27f00-172b-48b5-820d-9b3c59a7160d

core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosStatusDialog.kt
biplab1, line 1: Can you tell me why the core UI components was changed? Core components shouldn't be touched unless there is a global change required wherever the component is used.

core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosStatusDialog.kt
```diff
@@ -57,37 +59,43 @@ fun MifosStatusDialog(
...
 
-    Box(
-        modifier.fillMaxSize(),
-        contentAlignment = Alignment.Center,
+    if (status == ResultStatus.SUCCESS) {
+        LaunchedEffect(Unit) {
+            delay(1500)
+            onConfirm()
+        }
+    }
```
biplab1: The delay should be added at the feature level and not at the core level. Any customization required for a particular outcome should be handled at the feature level unless we require that change everywhere, in which case, the core components can be updated. Please review.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientDetailsProfile/ClientProfileDetailsViewModel.kt
```diff
@@ -62,6 +62,15 @@ internal class ClientProfileDetailsViewModel(
 
     init {
         getClientAndObserveNetwork()
+        observeGlobalUpdates()
+    }
+
+    private fun observeGlobalUpdates() {
```
biplab1: Can you tell me thought behind using this name `observerGlobalUpdates`. It doesn't clearly say anything about the kind of update it is handling.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientEditProfile/ClientProfileEditScreen.kt
```diff
@@ -260,5 +266,31 @@ private fun ClientProfileEditDialogs(
...
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .padding(DesignToken.padding.large),
+                ) {
+                    MifosStatusDialog(
+                        status = state.dialogState.status,
+                        onConfirm = {
+                            onAction(ClientProfileEditAction.OnNext)
+                        },
+                        btnText = stringResource(Res.string.dialog_continue),
```
biplab1, line 299: Please follow the naming convention for the string resources. The naming convention of the string resource should be of the following format:

`[feature_name]_[string_resource_name]`

Say, we are working with client charges and the string resource is "No charges available". Then the name of string resource should be:

`client_charges_no_charges_available`

biplab1 (changes requested):
Please address the requested changes below.

core/ui/src/commonMain/kotlin/com/mifos/core/ui/components/MifosStatusDialog.kt
kartikey004: > Can you tell me why the core UI components was changed? Core components shouldn't be touched unless there is a global change required wherever the component is used.

I updated the core component because I thought the auto-closing dialog as a global UX improvement rather than just for this specific screen.

now I understand the requirement to keep core components purely presentational. I will add a `showButton` parameter instead allowing the feature layer to decide the behavior without enforcing it globally.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientDetailsProfile/ClientProfileDetailsViewModel.kt
kartikey004: Initially I thought that maybe in the future we can use this function to update clients from anywhere in the application, from any screen. hence kept it as `observeGlobalUpdates`.

I'll rename it, I think `observeClientUpdates` is clearer.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientEditProfile/ClientProfileEditScreen.kt
kartikey004: I used `dialog_continue` because it was previously defined in the strings.xml. I reused the existing resource.

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientEditProfile/ClientProfileEditScreen.kt
```diff
@@ -260,5 +269,38 @@ private fun ClientProfileEditDialogs(
...
+                        successTitle = stringResource(Res.string.client_identifiers_identities_success_text),
+                        successMessage = state.dialogState.msg,
+                        failureTitle = stringResource(Res.string.client_identifiers_error_text),
+                        failureMessage = state.dialogState.msg,
+                        showButton = state.dialogState.status == ResultStatus.FAILURE
+                    )
+                }
+            }
+        }
```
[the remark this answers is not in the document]
kartikey004: should i do `dismissOnBackPress = false` and `dismissOnClickOutside = false` for both states?

feature/client/src/commonMain/kotlin/com/mifos/feature/client/clientEditProfile/ClientProfileEditScreen.kt
```diff
@@ -14,9 +14,12 @@ import androidclient.feature.client.generated.resources.account_number_prefix
 import androidclient.feature.client.generated.resources.arrow_up
 import androidclient.feature.client.generated.resources.cancel
 import androidclient.feature.client.generated.resources.choose_from_option
+import androidclient.feature.client.generated.resources.client_identifiers_error_text
+import androidclient.feature.client.generated.resources.client_identifiers_identities_success_text
```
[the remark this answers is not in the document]
kartikey004:
```
   btnText = stringResource(Res.string.client_profile_edit_dialog_continue),
                        successTitle = stringResource(Res.string.client_profile_edit_success_text),
                        successMessage = state.dialogState.msg,
                        failureTitle = stringResource(Res.string.client_profile_edit_error_text)
```

is this correct naming?

biplab1 (approved):

PR #2517, opened 2025-10-04
Feature/mifosac 579 client fields

SnehaghoshBarsha444:
Fixes - [Jira-#Issue_Number->MIFOSAC-579](https://mifosforge.jira.com/browse/MIFOSAC-579)

Please make sure these boxes are checked before submitting your pull request - thanks!

- [✔️] Run the static analysis check `./gradlew check` or `ci-prepush.sh` to make sure you didn't break anything

- [✔️] If you have multiple commits please combine them into one commit by squashing them.

Issue #579, opened 2017-03-14
Progress Bar in offline mode won't stop loading


aj019:
**Summary:** 

The progress bar which is shown before loading the client image never stops loading in offline mode.

**Steps to reproduce:** 

Open the app turn on offline mode.
Now search for a client and click on the client.
You will be taken to the client activity but the progress bar that shows the client image won't stop loading.

**Expected behavior:** 

The app should load the clients image and if it is not able to load the client image it should load a default image.

**Observed behavior:** 

The progress bar keeps loading without displaying any image.

**Device and Android version:** 

Nexus 5s API 23

**Screenshots:** 

![ezgif com-video-to-gif 2](https://cloud.githubusercontent.com/assets/7893859/24431991/d544fa7a-143c-11e7-9977-2c4acaa3f7d3.gif)

karsai1993:
Hi @aj019 ! Do you plan to solve it? I would try it. What do you type when you search for a client? When I type something I get the error: Failed to fetch resources.

ishan1604:
@karsai1993 Either try searching with any name or any number (it is the client ID)

karsai1993:
@ishan1604 I get search results when my device is connected to the internet. Now, I can reproduce the issue. However, when I do not turn on offline mode, searching and clicking to a client, I face the same issue: image does not get loaded, loading does not stop and additionally I get a fetching error: Client not found.
So, this issue is worth being extended to online mode, isn't it?

Apart from this, could you please tell me what the difference should be between offline mode (my device is not connected to the internet) and special mode(my device is connected to the internet but offline mode is turned on)? Or no matter what, should my device always be connected to the internet?

ishan1604:
@karsai1993 

I am not sure If I am following this correctly. But we are just talking about progress bar not disappearing in case image fails to load - in this issue.

Client not found error should be tackled in a separate issue not here. And if you get it because of the image then it is wrong and must be fixed.

I think what the issue creator has suggested should be done.

For offline mode you'll have to take the discussion to Gitter Channel or Mailing List. Since it is a Mifos Specific case and they can tell you about it better. I'd like to keep this space clean :)

karsai1993:
Thanks for answer @ishan1604 !

aj019:
@therajanmaurya @ishan1604 I would like to work on this issue.

aj019:
I have submitted a PR https://github.com/openMF/android-client/pull/592 for this issue. Please take a look @therajanmaurya @ishan1604.

[...]

PR #2578, opened 2026-01-17
fix: add missing gender, clientType and clientClassification in ClientProfileDetails screen

amanna13:
Fixes - [Jira-#579](https://mifosforge.jira.com/browse/MIFOSAC-579)

- Created separate database entities for gender, client type, and client classification following the same pattern as ClientStatusEntity
- Added `ClientGenderEntity`, `ClientTypeEntity`, `ClientClassificationEntity` in core/database
- Register converters in the `CustomTypeConverters` for serializing/deserializing these entities to/from JSON
- Added `GetGenderOptions`, `GetClientTypeOptions`, `GetClientClassificationOptions` in `core/network` and updated the `ClientMapper` accordingly.  
- Updated ClientProfileDetailsViewModel to access data properly instead of hardcoded "Not Available" fallback

>[!Note]
`GenderOptions`, `ClientTypeOptions`, `ClientClassificationOptions.kt` though present in the `core/model/objects/organisations` was never used. For consistency we have added new models in the `core/network` and kept the old files as is.

|image | image |

### Also did respective changes in the ClientEditDetailsScreen
image
image

- [x] Run the static analysis check `./gradlew check` or `ci-prepush.sh` to make sure you didn't break anything

> Related PR - #2517

## Summary by CodeRabbit

* **New Features**
  * Added client gender, client type, and client classification fields across profiles and edit screens.
  * Profiles now display these attributes with "not available" fallbacks when missing.
  * Edit screens populate and enable dropdowns for these fields and include them in data sync.
  * Dropdown options are driven by available server-provided lists for each field.

### DIFF
--- a/core/database/src/commonMain/kotlin/com/mifos/room/entities/client/ClientClassificationEntity.kt
+++ b/core/database/src/commonMain/kotlin/com/mifos/room/entities/client/ClientClassificationEntity.kt
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2025 Mifos Initiative
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * See https://github.com/openMF/android-client/blob/master/LICENSE.md
+ */
+package com.mifos.room.entities.client
+
+import com.mifos.core.model.utils.Parcelable
+import com.mifos.core.model.utils.Parcelize
+import com.mifos.room.utils.Entity
+import com.mifos.room.utils.PrimaryKey
+import kotlinx.serialization.Serializable
+
+@Parcelize
+@Serializable
+@Entity(
+    indices = [],
+    inheritSuperIndices = false,
+    primaryKeys = [],
+    foreignKeys = [],
+    ignoredColumns = [],
+    tableName = "ClientClassification",
+)
+data class ClientClassificationEntity(
+    @PrimaryKey(autoGenerate = true)
+    val id: Int = 0,
+
+    val name: String? = null,
+) : Parcelable
--- a/core/network/src/commonMain/kotlin/com/mifos/core/network/model/GetClientClassificationOptions.kt
+++ b/core/network/src/commonMain/kotlin/com/mifos/core/network/model/GetClientClassificationOptions.kt
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2025 Mifos Initiative
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * See https://github.com/openMF/android-client/blob/master/LICENSE.md
+ */
+package com.mifos.core.network.model
+
+import kotlinx.serialization.Serializable
+
+/**
+ *
+ *
+ * @param id
+ * @param name
+ */
+
+@Serializable
+data class GetClientClassificationOptions(
+
+    val id: Long? = null,
+
+    val name: String? = null,
+)
--- a/core/network/src/commonMain/kotlin/com/mifos/core/network/model/GetClientClassificationOptions.kt
+++ b/core/network/src/commonMain/kotlin/com/mifos/core/network/model/GetClientClassificationOptions.kt
@@ -11,13 +11,6 @@ package com.mifos.core.network.model
 
 import kotlinx.serialization.Serializable
 
-/**
- *
- *
- * @param id
- * @param name
- */
-
 @Serializable
 data class GetClientClassificationOptions(
 
--- a/core/network/src/commonMain/kotlin/com/mifos/core/network/model/GetClientTypeOptions.kt
+++ b/core/network/src/commonMain/kotlin/com/mifos/core/network/model/GetClientTypeOptions.kt
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2025 Mifos Initiative
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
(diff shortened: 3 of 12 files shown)


biplab1:
@amanna13 In the edit screen, when the gender and classification types are not available in the back-end, are dropdowns disabled automatically?

core/database/src/commonMain/kotlin/com/mifos/room/entities/client/ClientClassificationEntity.kt
biplab1, line 1: You may remove `@Parcelize` and `Parcelable` only from the entities you created, since it is android specific and it will be eventually removed from all the data classes.

amanna13:
> @amanna13 In the edit screen, when the gender and classification types are not available in the back-end, are dropdowns disabled automatically?

@biplab1 For gender the dropdown is available even if its not available.
```
            clientTemplate.genderOptions?.let { list ->
                MifosTextFieldDropdown(
                    enabled = list.isNotEmpty(),
                    value = gender,
                    onValueChanged = { gender = it },
........
```
but for `clientType` and `classificationType` its disabled automatically if those are not available. 
```
            clientTemplate.clientTypeOptions?.let { list ->
                MifosTextFieldDropdown(
                    enabled = clientType.isNotEmpty(),
                    value = clientType,
                    onValueChanged = { clientType = it },
```
https://github.com/user-attachments/assets/1fbe75b5-44af-4574-90ad-59b644f766bc

You can have a look into this. Though these were there previously, I didn't touch these section in the code.

core/database/src/commonMain/kotlin/com/mifos/room/entities/client/ClientClassificationEntity.kt
amanna13: `ClientEntity.kt` the parent entity still uses `@Parcelize` and `Parcelable`. So I think it must be kept or else it ends up showing error.

How should i proceed with ? Or we can refactor it later on.

biplab1:
I had checked it earlier, after I added gender and client classification codes in the back-end via web-app, the Gender drop-down menu was working and Client Classification was still not working, the last one is what you have fixed.

So, currently make sure that both the cases are handled properly:
1. If Gender and Client Classification are available, then drop-down menu should be enabled and
2. If they are not setup in the back-end they should show a disabled drop-down menu.

core/database/src/commonMain/kotlin/com/mifos/room/entities/client/ClientClassificationEntity.kt
biplab1: Okay, then it can be refactored later.

biplab1:
Codes should be sorted by `Position` which is one of the fields (which is missing from the data class you created), and if there is a tie then use name for sorting. Can you update this too? Here is a screenshot from the web-app:

Client Classification:

Client Classification Codes

Gender:

Gender Codes


`Position` takes integers both negative and positive as well as zero. Make sure you consume all the fields in the API response (Gender and Client Classification) even though some of them are not used.

biplab1:
Can you check the order of the codes sent in the API response? I believe the back-end sorts the codes based on `position` field and no sorting would be required in front-end. 

Also, make sure the API responses are completely consumed by the data classes even if those fields are not used in the app currently.

amanna13:
@biplab1 yes, you are right. I checked it, the back-end automatically sorts the codes based on `position `field. Currently the frontend again adjusts it based on the `id`. So should I go ahead and make the changes.

biplab1:
@amanna13 Yes, please go ahead, the sorting code in the front-end is redundant.

amanna13:
> Also, make sure the API responses are completely consumed by the data classes even if those fields are not used in the app currently.

@biplab1 I have tested with postman, The API response returns this below  (for `client id = 3`)

```
    "gender": {
        "id": 28,
        "name": "Male"
    },
    "clientType": {
        "id": 24,
        "name": "Mobile"
    },
    "clientClassification": {
        "id": 34,
        "name": "Student"
    },
```

Notice that these don't have those extra fields configured yet - `position`, `description`, `active`, `mandatory`

biplab1:
Understood,. Can you push the latest updates so that I can do a final review?

amanna13:
@biplab1 Updated screenshot attached below (This will be in the case when the options are not configured in the backend)

https://github.com/user-attachments/assets/3cf2c1aa-1be4-4588-ae3d-5a92a4e1d840

When the options are configured 

https://github.com/user-attachments/assets/821e4ca4-c34d-49ab-abb5-1df524a094f0

core/network/src/commonMain/kotlin/com/mifos/core/network/model/GetClientClassificationOptions.kt
```diff
@@ -0,0 +1,27 @@
...
+package com.mifos.core.network.model
+
+import kotlinx.serialization.Serializable
+
+/**
+ *
+ *
+ * @param id
+ * @param name
+ */
```
biplab1: Please remove such comments unless very important.

core/network/src/commonMain/kotlin/com/mifos/core/network/model/GetClientTypeOptions.kt
```diff
@@ -0,0 +1,28 @@
...
+package com.mifos.core.network.model
+
+import kotlinx.serialization.Serializable
+
+/**
+ *
+ *
+ * @param id
+ * @param name
+ */
```
biplab1: Same here, please review.

core/network/src/commonMain/kotlin/com/mifos/core/network/model/GetGenderOptions.kt
```diff
@@ -0,0 +1,28 @@
...
+package com.mifos.core.network.model
+
+import kotlinx.serialization.Serializable
+
+/**
+ *
+ *
+ * @param id
+ * @param name
+ */
```
biplab1: Same here, please review.

biplab1 (approved):
Some minor changes requested. 

core/network/src/commonMain/kotlin/com/mifos/core/network/model/GetClientClassificationOptions.kt
amanna13: Actually in the other data classes inside the same package it was there. Hence I followed the same format

PR #2589, opened 2026-01-29
Fix cross-graph navigation crash by hoisting events to root controller

Kartikey15dem:
Fixes - [Jira-#MIFOSAC-638](https://mifosforge.jira.com/browse/MIFOSAC-638)

## Description
This PR fixes a runtime crash occurring when attempting to navigate from the Client/Loan screen (inside the authenticated tabs) to the Data Table screen.

## The Problem (Root Cause)
The app uses a **nested navigation structure**:
1. **Root NavHost:** Manages global screens like `AuthenticatedNavbar` and `DataTable`.
2. **Nested NavHost:** Manages the tabs inside `AuthenticatedNavbar` (including `ClientNavGraph`).

The crash (`java.lang.IllegalArgumentException`) occurred because we were triggering navigation to `DataTableRoute` using the **nested** `NavController`. Since `DataTableRoute` is only registered in the **Root** `NavHost`, the nested controller could not resolve the destination and crashed.

### Behavior Changes
| Before | After |
| :---: | :---: |
| <video src="https://github.com/user-attachments/assets/cf479e6b-6228-4102-8367-7573d499da26"  controls height="150"></video> | <video src="https://github.com/user-attachments/assets/ccecf0fb-d182-4b36-9cd7-d7915243f90b" controls height="150"></video> |

## The Solution
I applied the **Event Hoisting** pattern to bridge the two navigation graphs:
1. Updated `clientNavGraph` to expose a lambda callback (`onMoreInfoClicked`) instead of accepting a `NavController` to perform the navigation directly.
2. Hoisted this event up through the UI layer (`AuthenticatedNavbarNavigationScreen`).
3. Connected the callback in `authenticatedNavbarGraph` (the parent), where the **Root** `NavController` is available.

## Changes
- Modified `authenticatedNavbarGraph` to accept `onMoreInfoClicked`.
- Passed `navController::navigateToDataTable` into the graph at the root level.
- Ensured navigation commands for global screens are now executed by the parent controller, not the child.

## Verification
- [x] Verified that clicking "More Info" on a Client successfully opens the Data Table screen without crashing.
- [x] Verified that back navigation from Data Table works correctly and preserves the tab state.

## Summary by CodeRabbit

* **New Features**
  * Added a "More info" action that lets the app open detailed data views from client and document screens.

* **Chores**
  * Propagated the new action through the authenticated and client navigation flows so "More info" works consistently across the app.


TheKalpeshPawar (approved):

biplab1 (approved):

PR #2599, opened 2026-02-03
fix(client): enable loan repayment navigation from client loan accounts screen

Kartikey15dem:
Fixes - [Jira-#MIFOSAC-640](https://mifosforge.jira.com/browse/MIFOSAC-640)

##Description

This PR fixes the broken navigation flow for loan repayments from the Client Loan Accounts screen and introduces a more flexible navigation mechanism to handle both full entity objects and individual loan IDs.

## Problem
1. **Empty Navigation Lambda**: The lambda responsible for navigating to the loan repayment screen from `ClientLoanAccountsScreen` was empty, making the "Make Repayment" action non-functional.
2. **Type Mismatch**: The existing navigation route required a `LoanWithAssociationsEntity`. While this works for the `LoanAccountSummaryScreen`, only the `loanId` (Int) is readily available in the context of the client loan list. 
3. **Dependency Constraints**: The `LoanWithAssociationsEntity` is deeply integrated into the current summary screen logic and could not be removed or replaced without breaking existing functionality.

### Behavior Changes
| Before | After |
| :---: | :---: |
| <video src="https://github.com/user-attachments/assets/0aaff9f1-900b-448f-a137-61d9526ff387" controls height="200"></video> | <video src="https://github.com/user-attachments/assets/0542972c-4087-49d1-896c-90ba8eb553f2" controls height="200"></video> |

## Solution & Implementation
To resolve this without breaking existing features, I implemented a dual-support navigation strategy:

### 1. Navigation Overloading
I overloaded the `navigateToLoanRepaymentScreen` function. It now supports:
- The original navigation using the full `LoanWithAssociationsEntity`.
- A new navigation path that only requires a `loanId: Int`.

### 2. Route & ViewModel Enhancements
- **Optional Route Parameters**: Updated `LoanRepaymentScreenRoute` to make fields like `loanAccountNumber` and `clientName` optional.
- **Data Fetching Logic**: Modified `LoanRepaymentViewModel` to handle cases where only a `loanId` is provided. If the loan details are missing on initialization, the ViewModel now uses the `LoanAccountSummaryRepository` to fetch the complete loan data by ID.
- **Action Update**: Updated `ClientLoanAccountsScreen` to correctly pass the `loanId` to the action handler.

## How I Managed the Constraints
Instead of refactoring the entire navigation system—which would have affected the `LoanAccountSummaryScreen`—I chose to **overload the navigation function**. This allows the app to "lazy-load" the necessary loan details in the `LoanRepaymentViewModel` if they weren't provided during navigation. This ensures a seamless user experience regardless of which screen the repayment flow is started from.

## Changes
- **`ClientNavigation.kt`**: Connected the navigation lambda to the repayment screen.
- **`LoanRepaymentScreenRoute.kt`**: Added `loanId` overload and default parameter values.
- **`LoanRepaymentViewModel.kt`**: Added logic to fetch loan details by ID if parameters are missing.
- **`ClientLoanAccountsScreen.kt` / `ViewModel.kt`**: Updated actions to pass `loanId`.

## Release Notes

* **Bug Fixes**
  * Corrected loan repayment process to properly identify the selected loan when initiating a repayment

* **Improvements**
  * Loan account details now automatically load when accessing the repayment screen, reducing manual data entry requirements
  * Simplified and optimized navigation flow for initiating loan repayments

### DIFF
--- a/feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentScreen.kt
+++ b/feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentScreen.kt
@@ -105,25 +105,34 @@ internal fun LoanRepaymentScreen(
     viewmodel: LoanRepaymentViewModel = koinViewModel(),
 ) {
     val uiState by viewmodel.loanRepaymentUiState.collectAsStateWithLifecycle()
+    val loanDetails by viewmodel.loanDetailsState.collectAsStateWithLifecycle()
 
     LaunchedEffect(key1 = Unit) {
-        viewmodel.checkDatabaseLoanRepaymentByLoanId()
+        if (loanDetails.loanAccountNumber.isNotEmpty()) {
+            viewmodel.checkDatabaseLoanRepaymentByLoanId()
+        }
     }
 
     LoanRepaymentScreen(
-        loanId = viewmodel.arg.loanId,
-        clientName = viewmodel.arg.clientName,
-        loanProductName = viewmodel.arg.loanProductName,
-        amountInArrears = viewmodel.arg.amountInArrears,
-        loanAccountNumber = viewmodel.arg.loanAccountNumber,
+        loanId = loanDetails.loanId,
+        clientName = loanDetails.clientName,
+        loanProductName = loanDetails.loanProductName,
+        amountInArrears = loanDetails.amountInArrears,
+        loanAccountNumber = loanDetails.loanAccountNumber,
         uiState = uiState,
         navigateBack = navigateBack,
-        onRetry = { viewmodel.checkDatabaseLoanRepaymentByLoanId() },
+        onRetry = {
+            if (loanDetails.loanAccountNumber.isEmpty()) {
+                viewmodel.loadLoanById()
+            } else {
+                viewmodel.checkDatabaseLoanRepaymentByLoanId()
+            }
+        },
         submitPayment = {
             viewmodel.submitPayment(it)
         },
         onLoanRepaymentDoesNotExistInDatabase = {
-            viewmodel.loanLoanRepaymentTemplate()
+            viewmodel.loadLoanRepaymentTemplate()
         },
         formatCurrency = viewmodel::formatCurrency,
         calculateTotal = viewmodel::calculateTotal,
--- a/feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentViewModel.kt
+++ b/feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentViewModel.kt
@@ -12,34 +12,90 @@ package com.mifos.feature.loan.loanRepayment
 import androidclient.feature.loan.generated.resources.Res
 import androidclient.feature.loan.generated.resources.feature_loan_failed_to_load_loan_repayment
 import androidclient.feature.loan.generated.resources.feature_loan_payment_failed
+import androidclient.feature.loan.generated.resources.feature_loan_profile_error_details_not_found
+import androidclient.feature.loan.generated.resources.feature_loan_profile_failed_to_load_loan
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import androidx.navigation.toRoute
 import com.mifos.core.common.utils.CurrencyFormatter
 import com.mifos.core.common.utils.DataState
+import com.mifos.core.data.repository.LoanAccountSummaryRepository
 import com.mifos.core.data.repository.LoanRepaymentRepository
 import com.mifos.room.entities.accounts.loans.LoanRepaymentRequestEntity
-import com.mifos.room.entities.templates.loans.LoanRepaymentTemplateEntity
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.launch
 import kotlin.math.round
 
 class LoanRepaymentViewModel(
     savedStateHandle: SavedStateHandle,
     private val repository: LoanRepaymentRepository,
+    private val summaryRepository: LoanAccountSummaryRepository,
 ) : ViewModel() {
 
-    val arg = savedStateHandle.toRoute<LoanRepaymentScreenRoute>()
+    private val args = savedStateHandle.toRoute<LoanRepaymentScreenRoute>()
+    private val _loanDetailsState = MutableStateFlow(LoanDetails())
+    val loanDetailsState = _loanDetailsState.asStateFlow()
 
     private val _loanRepaymentUiState =
         MutableStateFlow<LoanRepaymentUiState>(LoanRepaymentUiState.ShowProgressbar)
     val loanRepaymentUiState: StateFlow<LoanRepaymentUiState> get() = _loanRepaymentUiState
 
-    fun loanLoanRepaymentTemplate() {
+    init {
+        if (args.loanAccountNumber.isEmpty()) {
+            loadLoanById()
+        } else {
+            _loanDetailsState.value = _loanDetailsState.value.copy(
+                loanAccountNumber = args.loanAccountNumber,
+                loanId = args.loanId,
+                clientName = args.clientName,
+                loanProductName = args.loanProductName,
+                amountInArrears = args.amountInArrears,
+            )
+        }
+    }
+
+    fun loadLoanById() {
         viewModelScope.launch {
-            repository.getLoanRepayTemplate(arg.loanId).collect { state ->
+            summaryRepository.getLoanById(args.loanId).collect { dataState ->
+                when (dataState) {
+                    is DataState.Loading -> {
+                        _loanRepaymentUiState.value = LoanRepaymentUiState.ShowProgressbar
+                    }
+
+                    is DataState.Success -> {
+                        val loanWithAssociations = dataState.data
+                        if (loanWithAssociations == null) {
+                            _loanRepaymentUiState.value = LoanRepaymentUiState.ShowError(
+                                Res.string.feature_loan_profile_error_details_not_found,
+                            )
+                            return@collect
+                        }
+                        _loanDetailsState.value = _loanDetailsState.value.copy(
+                            loanId = loanWithAssociations.id,
+                            clientName = loanWithAssociations.clientName,
+                            loanProductName = loanWithAssociations.loanProductName,
+                            amountInArrears = loanWithAssociations.summary.totalOverdue,
+                            loanAccountNumber = loanWithAssociations.accountNo,
+                        )
+                        checkDatabaseLoanRepaymentByLoanId()
+                    }
+
+                    is DataState.Error -> {
+                        _loanRepaymentUiState.value = LoanRepaymentUiState.ShowError(
+                            Res.string.feature_loan_profile_failed_to_load_loan,
+                        )
+                    }
+                }
+            }
+        }
+    }
+
+    fun loadLoanRepaymentTemplate() {
+        viewModelScope.launch {
+            repository.getLoanRepayTemplate(args.loanId).collect { state ->
                 when (state) {
                     is DataState.Error ->
                         _loanRepaymentUiState.value =
@@ -47,13 +103,20 @@ class LoanRepaymentViewModel(
                                 Res.string
                                     .feature_loan_failed_to_load_loan_repayment,
                             )
+
                     DataState.Loading ->
                         _loanRepaymentUiState.value = LoanRepaymentUiState.ShowProgressbar
-                    is DataState.Success ->
-                        _loanRepaymentUiState.value =
-                            LoanRepaymentUiState.ShowLoanRepayTemplate(
-                                state.data ?: LoanRepaymentTemplateEntity(),
+
+                    is DataState.Success -> {
+                        val template = state.data
+                        if (template == null) {
+                            _loanRepaymentUiState.value = LoanRepaymentUiState.ShowError(
+                                Res.string.feature_loan_failed_to_load_loan_repayment,
                             )
+                            return@collect
+                        }
+                        _loanRepaymentUiState.value = LoanRepaymentUiState.ShowLoanRepayTemplate(template)
+                    }
                 }
             }
         }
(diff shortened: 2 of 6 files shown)


feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentViewModel.kt
```diff
@@ -12,32 +12,74 @@ package com.mifos.feature.loan.loanRepayment
...
+    val arg: StateFlow<LoanRepaymentScreenRoute> = _arg.asStateFlow()
 
     private val _loanRepaymentUiState =
         MutableStateFlow<LoanRepaymentUiState>(LoanRepaymentUiState.ShowProgressbar)
     val loanRepaymentUiState: StateFlow<LoanRepaymentUiState> get() = _loanRepaymentUiState
 
+    init {
+        if (_arg.value.loanAccountNumber.isEmpty()) {
+            loadLoanById()
+        }
```
itsPronay: Create a variable in the viremodel called `loanAccountNumber` and use it here, don't do `_arg.value.dataName` inside.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentScreen.kt
```diff
@@ -96,17 +96,20 @@ internal fun LoanRepaymentScreen(
     viewmodel: LoanRepaymentViewModel = koinViewModel(),
 ) {
     val uiState by viewmodel.loanRepaymentUiState.collectAsStateWithLifecycle()
+    val arg by viewmodel.arg.collectAsStateWithLifecycle()
 
     LaunchedEffect(key1 = Unit) {
```
sahilshivekar, line 110: This LaunchedEffect is making the code a bit less readable. It can be done in ViewModel itself.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentViewModel.kt
```diff
@@ -12,32 +12,74 @@ package com.mifos.feature.loan.loanRepayment
...
 
-    val arg = savedStateHandle.toRoute<LoanRepaymentScreenRoute>()
+    private val _arg = MutableStateFlow(savedStateHandle.toRoute<LoanRepaymentScreenRoute>())
+    val arg: StateFlow<LoanRepaymentScreenRoute> = _arg.asStateFlow()
 
     private val _loanRepaymentUiState =
         MutableStateFlow<LoanRepaymentUiState>(LoanRepaymentUiState.ShowProgressbar)
     val loanRepaymentUiState: StateFlow<LoanRepaymentUiState> get() = _loanRepaymentUiState
 
+    init {
```
sahilshivekar, line 46: Call checkDatabaseLoanRepaymentByLoanId() in this init block only. (and remove it from the loadLoanById function)

it would look like this 
```
init {
    if (_arg.loanAccountNumber.isEmpty()) {
        loadLoanById()
    }
    checkDatabaseLoanRepaymentByLoanId()
```

sahilshivekar (changes requested):

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentViewModel.kt
Kartikey15dem: Yes, I observed that checkDatabaseLoanRepaymentByLoanId() can be called regardless of the fact whether the loan details are being loaded or we already have from the args as in both we are getting the loanId by args but I have not called the function in viewmodel and kept it like it was before my changes because calling it only in init block may display stale data until the screen is in backstack . Suppose , if we navigate to another screen and navigate back to this screen then launched effect will load the data again but the init block will not do so until the viewmodel is destroyed and created again.

sahilshivekar: "If we navigate to another screen and navigate back to this screen then launched effect will load the data again"

Are you sure about this? because on the internet I see that LaunchedEffect runs only two times 1. Entering in the composition, 2. when the keys are updated. Navigating to other screen doesn't remove current screen from composition, current screen stays in the backstack. So navigating back will not add LaunchedEffect again in the composition and hence checkDatabaseLoanRepaymentByLoanId will not be called again I think.
Have you tested that? 

Because if that's not happening then we should add a button to load them

Kartikey15dem: @sahilshivekar 
I cross checked by adding an empty notification screen and navigated to that screen and then navigated back and the launched effect code ran again . You can check once for account 053 in maria. Although I am still a bit confused as different LLMs are giving different answers and clarify it from any senior but keeping the launched effect is better in either case as it manually checked it .

Kartikey15dem: You can check for loan account 053 in maria.

sahilshivekar: I am also confused that why LaunchedEffect is again triggering if the composition is in the backstack.
Need to ask about it to a senior. Ask about it in daily standup tmrw if you are also unsure

Kartikey15dem: @sahilshivekar 
But still I think it can be approved because using launchedeffect won't cause any harm, instead it will better.

sahilshivekar (approved):
LGTM👍

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentViewModel.kt
```diff
@@ -12,34 +12,75 @@ package com.mifos.feature.loan.loanRepayment
...
+                        _loanRepaymentUiState.value = LoanRepaymentUiState.ShowError(
+                            Res.string.feature_loan_unknown_error_occured,
+                        )
+                    }
+                }
+            }
+        }
+    }
+
     fun loanLoanRepaymentTemplate() {
```
biplab1: Shouldn't this be?

```suggestion
    fun loadLoanRepaymentTemplate() {
```

biplab1:
@coderabbitai Improve the PR title following this: https://github.com/openMF/mobile-wallet/wiki/Commit-style-guide. Just give the title.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentViewModel.kt
```diff
@@ -12,34 +12,75 @@ package com.mifos.feature.loan.loanRepayment
...
 import kotlin.math.round
 
 class LoanRepaymentViewModel(
     savedStateHandle: SavedStateHandle,
     private val repository: LoanRepaymentRepository,
+    private val summaryRepository: LoanAccountSummaryRepository,
 ) : ViewModel() {
 
-    val arg = savedStateHandle.toRoute<LoanRepaymentScreenRoute>()
+    private val _arg = MutableStateFlow(savedStateHandle.toRoute<LoanRepaymentScreenRoute>())
```
biplab1: Does it require to be a state flow? Please review.

Kartikey15dem: @biplab1  yes I made it state flow on purpose so that the LoanRepaymentScreen composable which is taking it as parameter recomposes after data is fetched in the case when we are navigating directly from ClientLoanAccountScreen without any data as in that case the args don't have the data to show .

biplab1: Navigation args from `SavedStateHandle.toRoute()` are static and should remain immutable. Mutating `_arg` mixes navigation arguments with UI state. I think it would be cleaner to keep `args` as a plain value and expose a separate `StateFlow` for loaded loan data.

Please refer other modules in the Client Flow.

biplab1: You should separate navigation arguments from UI state. For example:

Navigation args (immutable):

```kotlin
private val args = savedStateHandle.toRoute<LoanRepaymentScreenRoute>()
```

UI state:

```kotlin
private val _loanDetails = MutableStateFlow<LoanDetails?>(null)
val loanDetails = _loanDetails.asStateFlow()
```

Then update UI state after loading data:

```kotlin
_loanDetails.value = LoanDetails(
    loanId = loanWithAssociations.id,
    clientName = loanWithAssociations.clientName,
    loanProductName = loanWithAssociations.loanProductName,
    ...
)
```

Compose should recompose from `loanDetails`, not from navigation args.

Kartikey15dem: @biplab1  I have made the required changes , also note that to avoid the race condition between             checkDatabaseLoanRepaymentByLoanId() and loadLoanById() , I have made sure that checkDatabaseLoanRepaymentByLoanId()  is called after fetching the data from  loadLoanById() if data is not available from navigation args in both cases when the screen loads and for retrying.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentViewModel.kt
```diff
@@ -12,34 +12,85 @@ package com.mifos.feature.loan.loanRepayment
...
+    fun loadLoanById() {
+        viewModelScope.launch {
+            summaryRepository.getLoanById(args.loanId).collect { dataState ->
+                when (dataState) {
+                    is DataState.Loading -> {
+                        _loanRepaymentUiState.value = LoanRepaymentUiState.ShowProgressbar
+                    }
+
+                    is DataState.Success -> {
+                        val loanWithAssociations = dataState.data ?: LoanWithAssociationsEntity()
```
biplab1: Can you tell us the reason for using an empty object when the object is null? The UI might display empty loan details instead of showing an error. Please review.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentViewModel.kt
```diff
@@ -134,3 +185,10 @@ class LoanRepaymentViewModel(
         } && paymentType.isNotBlank()
     }
 }
+data class LoanDetails(
+    var clientName: String = "",
+    var loanId: Int = 0,
+    var loanAccountNumber: String = "",
+    var loanProductName: String = "",
+    var amountInArrears: Double? = 0.0,
+)
```
biplab1, line 210:
```suggestion
internal data class LoanDetails(
    val clientName: String = "",
    val loanId: Int = 0,
    val loanAccountNumber: String = "",
    val loanProductName: String = "",
    val amountInArrears: Double? = 0.0,
)
```

Use `val` to keep UI state immutable and `internal` to scope it within the feature module.

Kartikey15dem: @biplab1  can't use internal as it is being used in the loanDetails stateflow which is publicly exposed to UI .

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentViewModel.kt
```diff
@@ -12,48 +12,110 @@ package com.mifos.feature.loan.loanRepayment
...
+                            loanProductName = loanWithAssociations.loanProductName,
+                            amountInArrears = loanWithAssociations.summary.totalOverdue,
+                            loanAccountNumber = loanWithAssociations.accountNo,
+                        )
+                        checkDatabaseLoanRepaymentByLoanId()
+                    }
+
+                    is DataState.Error -> {
+                        _loanRepaymentUiState.value = LoanRepaymentUiState.ShowError(
+                            Res.string.feature_loan_unknown_error_occured,
```
biplab1: Can you please show a relevant error instead of just "unknown error"?

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentViewModel.kt
```diff
@@ -12,48 +12,110 @@ package com.mifos.feature.loan.loanRepayment
...
+                when (dataState) {
+                    is DataState.Loading -> {
+                        _loanRepaymentUiState.value = LoanRepaymentUiState.ShowProgressbar
+                    }
+
+                    is DataState.Success -> {
+                        val loanWithAssociations = dataState.data
+                        if (loanWithAssociations == null) {
+                            _loanRepaymentUiState.value = LoanRepaymentUiState.ShowError(
+                                Res.string.feature_loan_unknown_error_occured,
```
biplab1: Same here, please review.

biplab1 (changes requested):

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanRepayment/LoanRepaymentViewModel.kt
```diff
@@ -12,48 +12,111 @@ package com.mifos.feature.loan.loanRepayment
...
+                when (dataState) {
+                    is DataState.Loading -> {
+                        _loanRepaymentUiState.value = LoanRepaymentUiState.ShowProgressbar
+                    }
+
+                    is DataState.Success -> {
+                        val loanWithAssociations = dataState.data
+                        if (loanWithAssociations == null) {
+                            _loanRepaymentUiState.value = LoanRepaymentUiState.ShowError(
+                                Res.string.feature_loan_profile_error_details_not_found,
```
biplab1, line 72: The error is still vague. This needs to be handled in a separate PR and also this module needs to be migrated from MVVM to MVI.

biplab1 (approved):
I am merging this for now.

PR #2603, opened 2026-02-06
feat(loan): Loan Transaction Table

sahilshivekar:
[Jira-MIFOSAC-651](https://mifosforge.jira.com/browse/MIFOSAC-651)

|  image | <video width="250" height="600" alt="image" src="https://github.com/user-attachments/assets/84add93d-54e0-42c6-9deb-7aa0d79d68c4" /> |

## Summary by CodeRabbit

* **New Features**
  * Redesigned loan transactions from card-based to a scrollable table with breakdown columns and per-row actions (undo, view receipts, view journal entries, view details).
  * New reusable table-row UI component and new size tokens for small/medium/large table cells.
  * Added model fields to surface outstanding balance and manual reversal status.

* **Localization**
  * Added string resources for table headers and transaction actions.


feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsScreen.kt
```diff
@@ -132,7 +139,27 @@ internal fun LoanTransactionsScreen(
...
+                            manuallyReversed = uiState.selectedRow?.manuallyReversed ?: false,
+                            onDismissRequest = { viewModel?.dismissBottomSheet() },
+                            onAction = { action ->
+                                viewModel?.onActionSelected(
+                                    action,
+                                    uiState.selectedRow?.id ?: "",
+                                )
+                            },
+                        )
                     }
```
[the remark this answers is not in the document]
sahilshivekar: isBottomSheetOpen & selectedRow are updated correctly. when isBottomSheetOpen is true there will be always a selectedRow

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsViewModel.kt
```diff
@@ -34,14 +36,90 @@ class LoanTransactionsViewModel(
...
+                        }
+
                     _loanTransactionsUiState.value =
                         LoanTransactionsUiState.ShowLoanTransaction(
-                            state.data,
+                            transactionsTableData = LoanTransactionsUiState.LoanTransactionsTableData(
+                                transactions = transactionsData,
+                            ),
                         )
+                }
```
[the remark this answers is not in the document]
sahilshivekar: CurrencyFormatter.format accepts null values for those parameters

core/model/src/commonMain/kotlin/com/mifos/core/model/objects/account/loan/Transaction.kt
```diff
@@ -41,4 +41,8 @@ data class Transaction(
     var penaltyChargesPortion: Double? = null,
 
     var overpaymentPortion: Double? = null,
+
+    var outstandingLoanBalance: Double? = null,
+
+    var manuallyReversed: Boolean? = null,
```
gurnoorpannu, line 47: These nullable fields are directly consumed by UI. Should we consider either Non-null defaults at model level  or Explicit null handling utilities to avoid formatting null values in the UI layer?
This would help prevent silent UI inconsistencies.

core/designsystem/src/commonMain/kotlin/com/mifos/core/designsystem/component/MifosTableRow.kt
```diff
@@ -0,0 +1,66 @@
...
+    },
+) {
+    Column(
+        modifier = modifier
+            .fillMaxWidth()
+            .background(backgroundColor),
+    ) {
+        Row(
+            modifier = Modifier
+                .horizontalScroll(scrollState),
```
gurnoorpannu: Nice approach here , Since MifosTableRow takes an external ScrollState,
just wanted to double-check that the same instance is shared between the
table header and all rows at the call site.
Using separate scroll states could cause the columns to go out of sync,
so confirming this is intentional.

core/model/src/commonMain/kotlin/com/mifos/core/model/objects/account/loan/Transaction.kt
sahilshivekar: Actually the API always sends the outstandingLoanBalance, manuallyReversed for each type of transaction.
I kept them nullable in the model because the other fields were also written as nullable.
The ViewModel will show "-" in case if it's null

core/designsystem/src/commonMain/kotlin/com/mifos/core/designsystem/component/MifosTableRow.kt
sahilshivekar: Yes all rows use same ScrollState to sync

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsScreen.kt
```diff
@@ -129,10 +146,27 @@ internal fun LoanTransactionsScreen(
...
+                    if (uiState.isBottomSheetOpen) {
+                        TransactionActionsBottomSheet(
+                            transactionType = uiState.selectedRow?.transactionType
+                                ?: TransactionType.UNKNOWN,
+                            manuallyReversed = uiState.selectedRow?.manuallyReversed ?: false,
+                            onDismissRequest = onDismissBottomSheet,
+                            onAction = { action ->
+                                uiState.selectedRow?.id?.let { id ->
+                                    onTransactionActionClick(action, id.toInt())
+                                }
```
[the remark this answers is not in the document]
sahilshivekar: that case won't happen

gurnoorpannu (approved):

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsScreen.kt
```diff
@@ -129,10 +146,27 @@ internal fun LoanTransactionsScreen(
                 }
 
                 is LoanTransactionsUiState.ShowLoanTransaction -> {
-                    if (uiState.loanWithAssociations.transactions.isEmpty()) {
+                    if (uiState.transactionsTableData!!.transactions.isEmpty()) {
```
kartikey004: @sahilshivekar is the force unwrap issue resolved?

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsScreen.kt
```diff
@@ -145,258 +179,341 @@ internal fun LoanTransactionsScreen(
...
 
+    val textCells: List<@Composable () -> Unit> = textValues.mapIndexed { index, value ->
+        {
             Box(
                 modifier = Modifier
-                    .fillMaxWidth()
-                    .padding(top = 12.dp, bottom = 8.dp),
+                    .fillMaxSize()
+                    .background(
+                        if (index in 6..10) {
```
kartikey004, line 389: @sahilshivekar hardcoded indices can be problematic in future i think. please address this if you have not.

kartikey004 (changes requested):

kartikey004:
@sahilshivekar rest LGTM. 👍

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsScreen.kt
sahilshivekar: @kartikey004 It won't be problematic in future as we can update the indices just in case if we add another column in the section. Actually it needs to be in that way, because we can't decide whether to color that cell or not based on it's value.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsScreen.kt
kartikey004: If `transactionsTableData` happens to be null for any reason, `!!` will cause an app crash.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsScreen.kt
kartikey004: alright

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsScreen.kt
sahilshivekar: @kartikey004 
actually transactionsTableData will always be non-null as the ShowLoanTransaction uiState is triggering after only after loading the transactions. (see it on line 101 in LoanTransactionsViewModel)

kartikey004 (approved):

techsavvy185:
@sahilshivekar I think it looks good, but I think you can make one more addition, you can add a spacer on the sides of the table so that it looks like a card that starts and also a spacer on the right side such that you can see the spacer after you scroll completely to the right and round the corners of the table. This way the heading won't look very close to the edge of the screen.

sahilshivekar:
> @sahilshivekar I think it looks good, but I think you can make one more addition, you can add a spacer on the sides of the table so that it looks like a card that starts and also a spacer on the right side such that you can see the spacer after you scroll completely to the right and round the corners of the table. This way the heading won't look very close to the edge of the screen.

I did exactly that initially but Rajan sir told me to remove it during daily standup

techsavvy185:
@sahilshivekar No, I'm not talking about that, I'm talking about 2 spacer in a row with the table not padding such that it feels like a card sliding into frame.

biplab1:
@sahilshivekar I think we can get rid of the Actions column, instead clicking anywhere in a row should bring up the Actions bottom sheet.

sahilshivekar:
> @sahilshivekar No, I'm not talking about that, I'm talking about 2 spacer in a row with the table not padding such that it feels like a card sliding into frame.

I got you now, nice idea

sahilshivekar:
> @sahilshivekar I think we can get rid of the Actions column, instead clicking anywhere in a row should bring up the Actions bottom sheet.

@biplab1 I have removed the Actions column and now clicking on the row shows the bottom sheet

sahilshivekar:
> @sahilshivekar No, I'm not talking about that, I'm talking about 2 spacer in a row with the table not padding such that it feels like a card sliding into frame.

@techsavvy185 I have implemented this, check it

techsavvy185:
@sahilshivekar It actually looks really good to me.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsScreen.kt
```diff
@@ -129,10 +141,27 @@ internal fun LoanTransactionsScreen(
                 }
 
                 is LoanTransactionsUiState.ShowLoanTransaction -> {
-                    if (uiState.loanWithAssociations.transactions.isEmpty()) {
+                    if (uiState.transactionsTableData!!.transactions.isEmpty()) {
```
biplab1: Please avoid the use of `!!`.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsViewModel.kt
```diff
@@ -34,14 +36,96 @@ class LoanTransactionsViewModel(
...
+                    val loanWithAssociations = state.data
+                    val currencyCode = loanWithAssociations.currency.code
+                    val maxDigits = loanWithAssociations.currency.decimalPlaces
+
+                    val transactionsData =
+                        loanWithAssociations.transactions.mapIndexed { index, transaction ->
+
+                            LoanTransactionsUiState.LoanTransactionsTableData.TransactionRowData(
+                                number = (index + 1).toString(),
+                                id = transaction.id?.toString() ?: "-",
```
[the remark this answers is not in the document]
biplab1: @sahilshivekar Please address this.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsScreen.kt
```diff
@@ -420,19 +520,24 @@ private class LoanTransactionsPreviewProvider : PreviewParameterProvider<LoanTra
...
-                        transaction,
-                        transaction,
-                        transaction,
-                        transaction,
-                        transaction,
-                        transaction,
-                    ),
+                transactionsTableData = LoanTransactionsUiState.LoanTransactionsTableData(
+                    transactions = List(10) { index ->
+                        val tx = transaction
```
biplab1: Please use a proper variable name.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsScreen.kt
sahilshivekar: This is fixed now

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanTransaction/LoanTransactionsScreen.kt
sahilshivekar: This is fixed now

biplab1 (approved):

PR #2618, opened 2026-02-20
fix: Loan repayment calendar showing wrong dates and status (MIFOSAC-705)

therajanmaurya:
## Summary

Fixes [MIFOSAC-705](https://mifosforge.jira.com/browse/MIFOSAC-705) - Repayment calendar for a new Loan Application shows wrong installment status and wrong dates.

### Problem
When creating a new Loan Application, the Schedule (Calendar) step was showing:
- Wrong dates (from an old/existing loan instead of the dates entered in Terms step)
- Wrong installment statuses (showing "paid"/"due" instead of "pending" for a NEW loan)

### Root Cause
In `NewLoanAccountViewModel.kt`, the `repaymentScheduler()` function was calling `syncLoanById(state.clientId)` which fetched an **existing loan** by client ID, instead of calculating a preview based on the form inputs the user just entered.

### Solution
- Added new API endpoint `POST /loans?command=calculateLoanSchedule` to calculate loan schedule preview without creating the loan
- Created `CalculateLoanScheduleUseCase` to handle schedule calculation
- Updated `repaymentScheduler()` to build `LoansPayload` from current form state and use the new API
- Schedule now correctly shows dates entered by user and all installments as "pending"

## Changes

| Layer | File | Change |
|-------|------|--------|
| Network | `LoanService.kt` | Added `calculateLoanSchedule` endpoint |
| Network | `DataManagerLoan.kt` | Added method to call API |
| Data | `LoanAccountRepository.kt` | Added interface method |
| Data | `LoanAccountRepositoryImp.kt` | Added implementation |
| Domain | `CalculateLoanScheduleUseCase.kt` | New use case file |
| Domain | `UseCaseModule.kt` | Registered use case in DI |
| Feature | `NewLoanAccountViewModel.kt` | Updated `repaymentScheduler()` logic |

## Test plan

- [ ] Create new loan application with specific dates in Terms step
- [ ] Verify Schedule step shows the correct dates from Terms
- [ ] Verify all installments show as "Pending" (not "Paid" or "Due")
- [ ] Test with client that has existing loans (should not show old loan data)
- [ ] Test different repayment frequencies (monthly, weekly)

[MIFOSAC-705]: https://mifosforge.jira.com/browse/MIFOSAC-705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Preview loan repayment schedules before creating the loan, showing a full repayment breakdown.

* **Refactor**
  * Loan workflow and UI updated to compute and display schedule previews from current form state.
  * Navigation and screen APIs updated to propagate account numbers through the apply-new-application and new-loan flows.

* **Model**
  * Loan currency representation adjusted (decimal places type changed), affecting schedule formatting.

### DIFF
--- a/core/network/src/commonMain/kotlin/com/mifos/core/network/services/LoanService.kt
+++ b/core/network/src/commonMain/kotlin/com/mifos/core/network/services/LoanService.kt
@@ -83,7 +83,7 @@ interface LoanService {
      * Used to preview the schedule before submitting the loan application.
      */
     @POST(APIEndPoint.CREATE_LOANS_ACCOUNTS + "?command=calculateLoanSchedule")
-    fun calculateLoanSchedule(@Body loansPayload: LoansPayload?): Flow<LoanWithAssociationsEntity>
+    fun calculateLoanSchedule(@Body loansPayload: LoansPayload?): Flow<HttpResponse>
 
     @GET(APIEndPoint.CREATE_LOANS_ACCOUNTS + "/template?templateType=individual")
     fun getLoansAccountTemplate(
--- a/core/network/src/commonMain/kotlin/com/mifos/core/network/services/LoanService.kt
+++ b/core/network/src/commonMain/kotlin/com/mifos/core/network/services/LoanService.kt
@@ -78,6 +78,13 @@ interface LoanService {
     @POST(APIEndPoint.CREATE_LOANS_ACCOUNTS)
     fun createLoansAccount(@Body loansPayload: LoansPayload?): Flow<HttpResponse>
 
+    /**
+     * Calculate loan repayment schedule without creating the loan.
+     * Used to preview the schedule before submitting the loan application.
+     */
+    @POST(APIEndPoint.CREATE_LOANS_ACCOUNTS + "?command=calculateLoanSchedule")
+    fun calculateLoanSchedule(@Body loansPayload: LoansPayload?): Flow<LoanWithAssociationsEntity>
+
     @GET(APIEndPoint.CREATE_LOANS_ACCOUNTS + "/template?templateType=individual")
     fun getLoansAccountTemplate(
         @Query("clientId") clientId: Int,
--- a/core/data/src/commonMain/kotlin/com/mifos/core/data/repository/LoanAccountRepository.kt
+++ b/core/data/src/commonMain/kotlin/com/mifos/core/data/repository/LoanAccountRepository.kt
@@ -10,9 +10,9 @@
 package com.mifos.core.data.repository
 
 import com.mifos.core.common.utils.DataState
+import com.mifos.core.model.objects.account.loan.RepaymentSchedule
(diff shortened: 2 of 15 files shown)


core/network/src/commonMain/kotlin/com/mifos/core/network/services/LoanService.kt
```diff
@@ -83,7 +83,7 @@ interface LoanService {
      * Used to preview the schedule before submitting the loan application.
      */
     @POST(APIEndPoint.CREATE_LOANS_ACCOUNTS + "?command=calculateLoanSchedule")
-    fun calculateLoanSchedule(@Body loansPayload: LoansPayload?): Flow<LoanWithAssociationsEntity>
```
therajanmaurya, line 86: it should have reponse type instead the of HttpResponse

Arinyadav1: So sir how can we handle this error msg

```json
{
  "developerMessage": "The request was invalid. This typically will happen due to validation errors which are provided.",
  "httpStatusCode": "400",
  "defaultUserMessage": "Validation errors exist.",
  "userMessageGlobalisationCode": "validation.msg.validation.errors.exist",
  "errors": [
    {
      "defaultUserMessage": "The parameter `numberOfRepayments` must be between 2 and 5.",
      "parameterName": "numberOfRepayments",
      "developerMessage": "The parameter `numberOfRepayments` must be between 2 and 5.",
      "userMessageGlobalisationCode": "validation.msg.loan.numberOfRepayments.is.not.within.expected.range",
      "args": [
        {
          "value": 60
        },
        {
          "value": 2
        },
        {
          "value": 5
        }
      ]
    }
  ]
}
```

which is I implemented like this in datamanager

```kotlin
fun calculateLoanSchedule(loansPayload: LoansPayload): Flow<RepaymentSchedule> {
        return mBaseApiManager.loanService.calculateLoanSchedule(loansPayload).map { response ->
            if (!response.status.isSuccess()) {
                val errorMessage = extractErrorMessage(response)

                throw IllegalStateException(errorMessage)
            }

            Json { ignoreUnknownKeys = true }.decodeFromString<RepaymentSchedule>(response.bodyAsText())
        }
    }
```

therajanmaurya: Look other repositories, we are already doing it

Arinyadav1: Sir this same approach use everywhere

Arinyadav1:
https://github.com/user-attachments/assets/b82ed024-5522-44b2-939c-bd3bca77a0df

PR #2629, opened 2026-03-02
feat(loan): add loan account general screen

amanna13:
Fixes - [Jira-#665](https://mifosforge.jira.com/browse/MIFOSAC-665)

Screenshot [Updated]

[loan account general](https://github.com/user-attachments/assets/34896c06-f7b1-4e4c-81a1-a43b9f7abe2f)

- [x] Run the static analysis check `./gradlew check` or `ci-prepush.sh` to make sure you didn't break anything

## Summary by CodeRabbit

* **New Features**
  * Added a Loan Account General screen with performance history, loan summary (principal, interest, fees, penalties) and detailed loan info (disbursement, purpose, officer, currency, amounts).
  * Added navigation and ViewModel to load and present loan-general data and integrated it into existing loan/profile navigation.
  * Added repository wiring to retrieve loan data for the new screen.
  * Added localized strings for summary and detail labels.


feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,473 @@
...
+                                        text = amount,
+                                        style = KptTheme.typography.bodySmall,
+                                        color = amtColors[i],
+                                    )
+                                }
+                            }
+                        }
+                    },
+                    widths = colWidths,
+                    backgroundColor = if (index % 2 != 0) {
```
sahilshivekar: Maybe keeping same background color for all rows (except header and footer row) will look better.

kartikey004 (changes requested):
@amanna13 could you please confirm if the font sizes used here are as intended? right now the headings feel slightly larger compared to other screens.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/navigation/LoanNavigation.kt
```diff
@@ -79,7 +81,17 @@ fun NavGraphBuilder.loanDestination(
...
+            when (item) {
+                LoanAccountProfileActionItem.General ->
+                    navController.navigateToLoanAccountGeneralScreen(loanId = loanId)
+                else -> Unit
+            }
+        },
+    )
+
+    loanAccountGeneralDestination(
+        navController = navController,
```
kartikey004: I think you should pull the latest changes and implement navigation using `ClientProfile` flow as reference.

kartikey004 (changes requested):

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,473 @@
...
+                    KptTheme.colorScheme.primary,
+                    0.3f,
+                ),
+                edgeOffset = DesignToken.padding.medium,
+                cornerShape = DesignToken.shapes.topMedium,
+            )
+
+            state.summaryRows.forEachIndexed { index, row ->
+                val amounts = listOf(row.original, row.paid, row.waived, row.writtenOff, row.outstanding, row.overDue)
+                val amtColors = listOf(textColor, KptTheme.colorScheme.primary, KptTheme.colorScheme.primary, KptTheme.colorScheme.primary, textColor, textColor)
```
sahilshivekar: move `amtColors` outside the loop

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,473 @@
...
+                        lerp(KptTheme.colorScheme.surface, KptTheme.colorScheme.primary, 0.08f)
+                    } else {
+                        KptTheme.colorScheme.surface
+                    },
+                    edgeOffset = DesignToken.padding.medium,
+                )
+            }
+
+            val totals = listOf(state.totalOriginal, state.totalPaid, state.totalWaived, state.totalWrittenOff, state.totalOutstanding, state.totalOverDue)
+            val totColors = listOf(textColor, KptTheme.colorScheme.primary, KptTheme.colorScheme.primary, KptTheme.colorScheme.primary, textColor, textColor)
```
sahilshivekar: since `totColors` and `amtColors` have same values, it will be better if we just keep a single generic variable for this

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,473 @@
...
+            MifosBreadcrumbNavBar(navController = navController)
+
+            Column(
+                modifier = Modifier
+                    .fillMaxSize()
+                    .verticalScroll(rememberScrollState()),
+            ) {
+                Text(
+                    text = stringResource(Res.string.feature_loan_general_section_performance_history),
+                    style = KptTheme.typography.labelLarge,
```
sahilshivekar, line 124: Other screens in the app mostly use Camel case for heading text.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,473 @@
...
+    Column(
+        modifier = modifier.fillMaxWidth(),
+        verticalArrangement = Arrangement.spacedBy(DesignToken.padding.medium),
+    ) {
+        MifosListingComponentOutline {
+            Column(
+                modifier = Modifier.fillMaxWidth(),
+                verticalArrangement = Arrangement.spacedBy(DesignToken.padding.medium),
+            ) {
+                MifosListingColumnItem(
```
sahilshivekar: You can add icons here and keep value of it in the same row as it's done in Loan Overview section of  `LoanAccountSummary` screen

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,473 @@
...
+                    modifier = Modifier.padding(horizontal = KptTheme.spacing.md),
+                )
+
+                Spacer(Modifier.height(KptTheme.spacing.xl))
+            }
+        }
+    }
+}
+
+@Composable
```
sahilshivekar: instead of adding MifosErrorComponent and MifosProgressIndicator in LoanAccountGeneralDialog and showing the screen content based on `if (state.dialogState == null)`, you can directly use when(state) in the start of `LoanAccountGeneralScreen` and keep states Loading, Error and Success to change what's displayed on screen.
Loan charge, loan disbursement, loan transaction screens use this approach.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/navigation/LoanNavigation.kt
amanna13: Yes, this is yet to be fixed.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
```diff
@@ -0,0 +1,260 @@
...
+                numberOfRepayments = loan.numberOfRepayments.toString(),
+                maturityDate = maturityDate,
+                disbursementDate = disbursementDate,
+                loanOfficer = loan.loanOfficerName.ifBlank { getString(Res.string.feature_loan_general_value_unassigned) },
+                currency = currencyDisplay,
+                loanPurpose = loan.loanPurposeName.ifBlank { getString(Res.string.feature_loan_general_value_not_available) },
+                externalId = loan.accountNo.ifBlank { getString(Res.string.feature_loan_general_value_not_available) },
+                proposedAmount = formatCurrency(loan.approvedPrincipal, currencyCode, decimalPlaces),
+                approvedAmount = formatCurrency(loan.approvedPrincipal, currencyCode, decimalPlaces),
+                disbursedAmount = formatCurrency(loan.principal, currencyCode, decimalPlaces),
```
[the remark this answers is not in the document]
sahilshivekar: @amanna13 There is a field `principal` also in `LoanWithAssociationsEntity`, maybe we need to ask Victor that `principal` is proposed amount or not

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
```diff
@@ -0,0 +1,260 @@
...
+            ""
+        }
+    }
+
+    private fun formatDateList(date: List<Int?>?): String {
+        if (date == null || date.size < 3) return ""
+        return Utils.getStringOfDate(date)
+    }
+
+    private fun formatCurrency(
```
sahilshivekar: I have actually seen this formatCurrency function in one more PR also, as we know for sure that `currencyCode`  will always have value then we don't need to check that here. 

you can just use following approach and it will work fine (as it's working in `LoanTransactionScreen`
CurrencyFormatter.format(
   <variable that holds Double?>,
    currencyCode,
    maxDigits,
)

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
```diff
@@ -0,0 +1,260 @@
...
+        val rowType: SummaryRowType,
+        val original: String,
+        val paid: String,
+        val waived: String,
+        val writtenOff: String,
+        val outstanding: String,
+        val overDue: String,
+    )
+
+    enum class SummaryRowType {
```
sahilshivekar: Why this enum is needed here? I don't see it affecting any logic on UI also

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
```diff
@@ -0,0 +1,260 @@
...
+                totalPaid = formatCurrency(summary.totalRepayment, currencyCode, decimalPlaces),
+                totalWaived = formatCurrency(summary.totalWaived, currencyCode, decimalPlaces),
+                totalWrittenOff = formatCurrency(summary.totalWrittenOff, currencyCode, decimalPlaces),
+                totalOutstanding = formatCurrency(summary.totalOutstanding, currencyCode, decimalPlaces),
+                totalOverDue = formatCurrency(summary.totalOverdue, currencyCode, decimalPlaces),
+            )
+        }
+    }
+
+    private fun formatActualDisbursementDate(date: List<Int?>?): String {
```
sahilshivekar: As date list will either be empty or have all the 3 Int (year, month, day), you can use following approach
```
transactionDate = 
if (transaction.date.isNotEmpty()) {
  DateHelper.getDateAsString(
      transaction.date,
  )
} else {
  "-"
},
```



also you should change `val actualDisbursementDate: List<Int?>? = null,` to `val actualDisbursementDate: List<Int>? = null,`

then you can remove these two date formatters and use same DateHelper at both places

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
amanna13: Refer to the design suggested in the ticket.

sahilshivekar: image

See this is from the ticket, maybe you are referring to the Web UI

amanna13: Ohh yes, I was referencing from the web ui. But isn't it looks much sleeker ?

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
amanna13: Can you please help me understand what icons are you referencing to  ?

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
amanna13: This was done, when each component had coloured indicators (refer to the design). But later it was confirmed that we dont need need such coloured indicators. Though I did rewired the UI, forgot to cleanup the viewmodel. Thanks for pointing !

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
sahilshivekar: Because some designs might look good to you and not to me and vice versa. I think you should ask about the designs & colors on channel to biplab and rajan sir as they are gonna finalize things.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
sahilshivekar: image

amanna13: I think this would clutter up the whole area, and neither this is present in the design (both app or web)

feature/loan/src/commonMain/composeResources/values/strings.xml
```diff
@@ -223,6 +223,38 @@
...
+    <string name="feature_loan_general_number_of_repayments">Number of Repayments</string>
+    <string name="feature_loan_general_maturity_date">Maturity Date</string>
+
+    <string name="feature_loan_general_summary_col_component">Component</string>
+    <string name="feature_loan_general_summary_col_original">Original</string>
+    <string name="feature_loan_general_summary_col_paid">Paid</string>
+    <string name="feature_loan_general_summary_col_waived">Waived</string>
+    <string name="feature_loan_general_summary_col_written_off">Written Off</string>
+    <string name="feature_loan_general_summary_col_outstanding">Outstanding</string>
+    <string name="feature_loan_general_summary_col_overdue">Over Due</string>
```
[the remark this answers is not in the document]
amanna13: It's present in the design as well as in the web app

core/data/src/commonMain/kotlin/com/mifos/core/data/repository/LoanAccountGeneralRepository.kt
```diff
@@ -0,0 +1,19 @@
...
+ */
+package com.mifos.core.data.repository
+
+import com.mifos.core.common.utils.DataState
+import com.mifos.room.entities.accounts.loans.LoanWithAssociationsEntity
+import kotlinx.coroutines.flow.Flow
+
+interface LoanAccountGeneralRepository {
+
+    fun getLoanById(loanId: Int): Flow<DataState<LoanWithAssociationsEntity?>>
```
niyajali: Why do repository methods return a room entity? Consider creating a domain model and mapping objects to translate data from the database into the domain model, which can then be returned for use in the feature module.

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/LoanAccountGeneralRepositoryImp.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.mifos.core.data.repository.LoanAccountGeneralRepository
+import com.mifos.core.network.datamanager.DataManagerLoan
+import com.mifos.room.entities.accounts.loans.LoanWithAssociationsEntity
+import kotlinx.coroutines.flow.Flow
+
+class LoanAccountGeneralRepositoryImp(
+    private val dataManagerLoan: DataManagerLoan,
+) : LoanAccountGeneralRepository {
+
+    override fun getLoanById(loanId: Int): Flow<DataState<LoanWithAssociationsEntity?>> {
```
niyajali: Similarly here..

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralNavigation.kt
```diff
@@ -0,0 +1,34 @@
...
+data class LoanAccountGeneralRoute(
+    val loanId: Int = -1,
+)
+
+fun NavGraphBuilder.loanAccountGeneralDestination(
+    navController: NavController,
+) {
+    composable<LoanAccountGeneralRoute> {
+        LoanAccountGeneralScreen(
+            navController = navController,
```
niyajali, line 27: why does the screen need a navController instance?

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,473 @@
...
+        state = state,
+        onAction = viewModel::trySendAction,
+    )
+}
+
+@Composable
+internal fun LoanAccountGeneralScreen(
+    state: LoanAccountGeneralState,
+    onAction: (LoanAccountGeneralAction) -> Unit,
+    navController: NavController,
```
niyajali, line 110: why being the navController accessed and passed around all the composables?

niyajali (changes requested):
check out mifos-pay project for referrence and update the PR accordingly with the required changes

core/data/src/commonMain/kotlin/com/mifos/core/data/repository/LoanAccountGeneralRepository.kt
amanna13: @niyajali If I'm not wrong, no other screens does this yet. It uses the same pattern as I did, to stay consistent I followed this approach

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralNavigation.kt
amanna13: the MifosBreadcrumNavBar needs the navcontroller. All other screens uses this, you can check the `LoanAccountSummaryScreen`, `ReccuringAccountScreen` `SavingsAccountScreen` andm more.

biplab1:
@amanna13 The UI doesn't seem to be aligned with the UI of the Client Flow like typography. Please review.

Performance history is a section and not the title of the screen. It should be Loan Account General. The mockup in ticket says General. However, we should add "Loan Account" as prefix since the Client Flow also has "General" screen as well. You may take a look at the client flow's General screen ensure consistency.

image

biplab1:
The table components are not being handled properly, please review. Like the "G" of Outstanding appears in the second line. Also the paddings on left and right end seem to be different.

Extreme Left:

Screenshot 2026-03-05 145652

Extreme Right:

Screenshot 2026-03-05 145702

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
```diff
@@ -0,0 +1,230 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    private suspend fun fillGeneralState(loan: LoanWithAssociationsEntity) {
+        val currencyCode = loan.currency.code
+        val maxDigits = loan.currency.decimalPlaces
+        val summary = loan.summary
```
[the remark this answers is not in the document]
amanna13: null safety fallback is not required as the backend assures that the data will be there/available

core/data/src/commonMain/kotlin/com/mifos/core/data/repository/LoanAccountGeneralRepository.kt
biplab1: @amanna13 I think we can start doing it from this module, take reference from `mifos-mobile` if required. Eventually we will be moving towards that only.

biplab1:
@amanna13 Please update the screen recording if UI has changed after the last changes.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,426 @@
...
+    }
+}
+
+@Composable
+private fun PerformanceHistoryRow(
+    label: String,
+    value: String,
+    modifier: Modifier = Modifier,
+    textStyle: TextStyle = TextStyle(
+        color = AppColors.customWhite,
```
biplab1: Have you tested the dark theme mode to check if this looks good?

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,426 @@
...
+        }
+    }
+}
+
+@Composable
+private fun LoanSummaryTableCell(
+    text: String,
+    style: TextStyle,
+    fontWeight: FontWeight,
+    textColor: androidx.compose.ui.graphics.Color,
```
biplab1: Please move this to import section: `androidx.compose.ui.graphics.Color`

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,426 @@
...
+@Composable
+private fun LoanSummaryTable(
+    state: LoanAccountGeneralState,
+    modifier: Modifier = Modifier,
+) {
+    val scrollState = rememberScrollState()
+    val textColor = KptTheme.colorScheme.onBackground
+    val componentWidth = DesignToken.sizes.tableCellWidthLarge
+    val amountWidth = DesignToken.sizes.tableCellWidthMedium
+    val colWidths = listOf(componentWidth, amountWidth, amountWidth, amountWidth, amountWidth, amountWidth, amountWidth)
```
biplab1: Until the dynamic table is implemented, please set the columns widths according to the content where too much extra space is not visible.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
```diff
@@ -0,0 +1,256 @@
...
+                        } else {
+                            mutableStateFlow.update {
+                                it.copy(dialogState = LoanAccountGeneralState.DialogState.Error(getString(Res.string.feature_loan_unknown_error_occured)))
+                            }
+                        }
+                    }
+
+                    is DataState.Error -> {
+                        mutableStateFlow.update {
+                            it.copy(dialogState = LoanAccountGeneralState.DialogState.Error(getString(Res.string.feature_loan_unknown_error_occured)))
```
biplab1: Can you please tell us the reason for throwing an unknown error instead of something specific?

amanna13: This has been fixed

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
amanna13: Currently it looks like this in dark theme mode. And this is almost same visually for most of the screens.
image

biplab1: The whole screen looks odd. Why are other parts not respecting the dark mode change? Several areas are still showing light background.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralNavigation.kt
```diff
@@ -0,0 +1,34 @@
...
+package com.mifos.feature.loan.loanAccountGeneral
+
+import androidx.navigation.NavController
+import androidx.navigation.NavGraphBuilder
+import androidx.navigation.compose.composable
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class LoanAccountGeneralRoute(
+    val loanId: Int = -1,
```
biplab1: Can you please tell us the reason for using `-1` as default value?

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
amanna13: I believe dark mode is still not fully implemented, it's broken. Not only this screen, each and every other screens have this same thing, light background in the middle between portion. I just implemented and followed what it was instead of fixing things out of a sudden.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralNavigation.kt
amanna13: Saw this pattern in `clientProfile` savings, recurring routes where `-1` is taken as the default as id's are supposed to be posiitve values

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
biplab1: Dark mode should work since we already migrated to KptTheme for ColorScheme.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
```diff
@@ -0,0 +1,256 @@
...
+                            try {
+                                fillGeneralState(loan)
+                            } catch (e: Exception) {
+                                mutableStateFlow.update {
+                                    it.copy(dialogState = LoanAccountGeneralState.DialogState.Error(e.message.toString()))
+                                }
+                            }
+                        } else {
+                            mutableStateFlow.update {
+                                it.copy(dialogState = LoanAccountGeneralState.DialogState.Error(getString(Res.string.feature_loan_unknown_error_occured)))
```
biplab1: Please use specific error instead of unknown.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
biplab1: I have created another ticket for investigating and fixing the dark mode issue. We can ignore it for this PR.

biplab1 (dismissed):

core/data/src/commonMain/kotlin/com/mifos/core/data/mappers/loan/LoanAccountGeneralMapper.kt
```diff
@@ -0,0 +1,183 @@
...
+import com.mifos.core.model.entity.loan.SavingAccountCurrency
+import com.mifos.room.entities.accounts.loans.ActualDisbursementDateEntity
+import com.mifos.room.entities.accounts.loans.LoanStatusEntity
+import com.mifos.room.entities.accounts.loans.LoanTimelineEntity
+import com.mifos.room.entities.accounts.loans.LoanTypeEntity
+import com.mifos.room.entities.accounts.loans.LoanWithAssociationsEntity
+import com.mifos.room.entities.accounts.loans.LoansAccountSummaryEntity
+import com.mifos.room.entities.accounts.savings.SavingAccountCurrencyEntity
+
+fun LoanWithAssociationsEntity.toModel(): LoanWithAssociations {
```
itsPronay: Why we are not using `com.mifos.core.network.data.AbstractMapper` ?

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/LoanAccountGeneralRepositoryImp.kt
```diff
@@ -0,0 +1,40 @@
...
+        return dataManagerLoan.getLoanById(loanId)
+            .asDataStateFlow()
+            .map { response ->
+                when (response) {
+                    is DataState.Loading -> DataState.Loading
+                    is DataState.Error -> DataState.Error(response.exception, response.data?.toModel())
+                    is DataState.Success -> DataState.Success(response.data?.toModel())
+                }
+            }
+            .flowOn(ioDispatcher)
```
itsPronay: should map before converting it to datastate flow. not after that

something like 

```kotlin
    override fun getLoanById(loanId: Int): Flow<DataState<LoanWithAssociations?>> {
        return dataManagerLoan.getLoanById(loanId)::mapToModel
            .asDataStateFlow()
            .flowOn(ioDispatcher)
```

core/model/src/commonMain/kotlin/com/mifos/core/model/entity/loan/LoanWithAssociations.kt
```diff
@@ -0,0 +1,180 @@
...
+    val loanPurposeName: String = "",
+    val loanOfficerId: Int = 0,
+    val loanOfficerName: String = "",
+    val loanType: LoanType = LoanType(),
+    val currency: SavingAccountCurrency = SavingAccountCurrency(),
+    val principal: Double = 0.0,
+    val approvedPrincipal: Double = 0.0,
+    val termFrequency: Int = 0,
+    val termPeriodFrequencyType: TermPeriodFrequencyType = TermPeriodFrequencyType(),
+    val numberOfRepayments: Int = 0,
```
itsPronay: are we supossed to have empty values?

core/model/src/commonMain/kotlin/com/mifos/core/model/entity/loan/LoanWithAssociations.kt
```diff
@@ -0,0 +1,180 @@
...
+data class SavingAccountCurrency(
+    val id: Int = 0,
+    val code: String? = null,
+    val name: String? = null,
+    val decimalPlaces: Int? = null,
+    val inMultiplesOf: Int? = null,
+    val displaySymbol: String? = null,
+    val nameCode: String? = null,
+    val displayLabel: String? = null,
+) : Parcelable
```
itsPronay: do not put all the classes in a single file. Do it separtely.  

create a package inside loan called loanwithAssociations and put all the files there.

core/model/src/commonMain/kotlin/com/mifos/core/model/entity/loan/LoanWithAssociations.kt
```diff
@@ -0,0 +1,180 @@
...
+    val year: Int? = null,
+    val month: Int? = null,
+    val date: Int? = null,
+) : Parcelable
+
+@Parcelize
+@Serializable
+data class LoansAccountSummary(
+    val loanId: Int? = null,
+    val currency: SavingAccountCurrency? = null,
```
itsPronay: use separate file

itsPronay (changes requested):

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
```diff
@@ -0,0 +1,256 @@
...
+        }
+    }
+
+    private fun observeNetwork() {
+        viewModelScope.launch {
+            networkMonitor.isOnline.collect { isConnected ->
+                mutableStateFlow.update { it.copy(networkConnection = isConnected) }
+            }
+        }
+    }
```
itsPronay: I dont see a valid reason to continuously observe network state. we should only check connectivity when making network requests....

Current behavior:
After the user fetches loan details, the data is displayed correctly. However, if the network connection is lost, all the displayed items disappear and a “no internet” screen is shown.

My question - why should users need an active internet connection to view or scroll through data that has already been fetched?>> 

we are already showing a snackbar to indicate loss of connectivity, so this additional thing impacts user experience. 
Only check the network state before making an API call and show a dialog /screen

itsPronay (changes requested):

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
amanna13: I’ve verified the codebase has mixed patterns: some screens observe continuously, while others do one-shot connectivity checks before making a API calls. For instance `ClientProfileGeneralViewModel` and some LoanAccounts screen observes continously 
Why haven't we yet standardize across modules, is there any specific reason we are using different approaches ?

Though I agree totally to the fact that there's no need to continuosly check for network state and I'm doing the necessary changes.

itsPronay: - I think most contributors implemented it by taking reference from Mifos Mobile, where we followed that approach.
- While reviewing PRs, I may have suggested to some contributors to use the one-shot approach, so they adopted that instead.

Hence, the difference. 

Yeah, you are right about it. We should standardise it.

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/LoanAccountGeneralRepositoryImp.kt
amanna13: I checked the repo pattern across modules. The common convention here is to wrap with asDataStateFlow() first, and then map if needed.

core/model/src/commonMain/kotlin/com/mifos/core/model/entity/loan/LoanWithAssociations.kt
amanna13: yes this kind of default-value pattern is used in other modules too. 

@biplab1 Can you also please verify and confirm.

core/data/src/commonMain/kotlin/com/mifos/core/data/mappers/loan/LoanAccountGeneralMapper.kt
amanna13: We’re not using `AbstractMapper` here because this mapper is for one-way only.
I checked that The existing `AbstractMapper` usage in the repo is for bidirectional mappings where both mapFromEntity and mapToEntity are meaningful

core/model/src/commonMain/kotlin/com/mifos/core/model/entity/loan/LoanWithAssociations.kt
Kartikey15dem: @amanna13  I checked mifos-mobile, it has the fields nullable and I think they can be null and it should be decided in the presentation layer how the null values should be handled.

core/data/src/commonMain/kotlin/com/mifos/core/data/mappers/loan/LoanAccountGeneralMapper.kt
Kartikey15dem: @amanna13  I think the same entity can be later used in bidirectional way and we can leave the modelToEntity mapping empty for now I guess.

amanna13: @Kartikey15dem Can you clarify ?

Kartikey15dem: @amanna13  I am just saying that keeping it bidirectional might be useful later and for now we can keep the domain to entity implementation empty for now  if we use abstract mapper.

core/model/src/commonMain/kotlin/com/mifos/core/model/entity/loan/LoanWithAssociations.kt
itsPronay: > yes this kind of default-value pattern is used in other modules too.
> 

It should not have a default value in the first place. Default values should be either null, an empty list, or something similar. 

I’m not sure which module you took this reference from, but please create a Jira ticket so we can address it later. 

Also, try to take references from the KMP template instead of other modules, as it is almost always up to date and we generally follow its structure.

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/LoanAccountGeneralRepositoryImp.kt
itsPronay: could you please tell which module you took reference from?

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
```diff
@@ -0,0 +1,266 @@
...
+    override fun handleAction(action: LoanAccountGeneralAction) {
+        when (action) {
+            LoanAccountGeneralAction.OnRetry -> loadLoanById()
+        }
+    }
+
+    private fun loadLoanById() {
+        loadJob?.cancel()
+        loadJob = viewModelScope.launch {
+            val isConnected = networkMonitor.isOnline.first()
```
niyajali: use kotlin `combine` operator function to combine the two streams into a single stream, and that way it will refeed the data from both sources when back online

niyajali (changes requested):
and the networkMontior could be moved in to the repository itself, and you can check the status before performing any action on other methods

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/LoanAccountGeneralRepositoryImp.kt
amanna13: Sorry for that, my mistake, I overlooked. I have fixed this now.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
```diff
@@ -0,0 +1,260 @@
...
+                        Res.string.feature_loan_general_detail_loan_purpose to loanPurpose,
+                        Res.string.feature_loan_general_detail_loan_officer to loanOfficer,
+                        Res.string.feature_loan_general_detail_currency to currencyDisplay,
+                    ),
+                    mapOf(
+                        Res.string.feature_loan_general_detail_proposed_amount to CurrencyFormatter.format(loan.approvedPrincipal, currencyCode, maxDigits),
+                        Res.string.feature_loan_general_detail_approved_amount to CurrencyFormatter.format(loan.approvedPrincipal, currencyCode, maxDigits),
+                        Res.string.feature_loan_general_detail_disbursed_amount to CurrencyFormatter.format(loan.principal, currencyCode, maxDigits),
+                    ),
+                ),
```
biplab1, line 194: One of the issues of using a Map is it is unstructured, so type safety could be a problem.
Instead of Map, we can use,
a proper model:
```Kotlin
data class DetailItem(
    val label: StringResource,
    val value: String
)
```
or even better:
```Kotlin
sealed interface LoanDetailItem
```

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
```diff
@@ -0,0 +1,260 @@
...
+                dialogState = null,
+                numberOfRepayments = loan.numberOfRepayments?.toString().orEmpty(),
+                maturityDate = maturityDate,
+                disbursementDate = disbursementDate,
+                loanOfficer = loanOfficer,
+                currency = currencyDisplay,
+                loanPurpose = loanPurpose,
+                externalId = externalId,
+                proposedAmount = CurrencyFormatter.format(loan.approvedPrincipal, currencyCode, maxDigits),
+                approvedAmount = CurrencyFormatter.format(loan.approvedPrincipal, currencyCode, maxDigits),
```
biplab1: How can both use the same field `approvedPrincipal`?

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralViewModel.kt
```diff
@@ -0,0 +1,260 @@
...
+                        Res.string.feature_loan_general_detail_approved_amount to CurrencyFormatter.format(loan.approvedPrincipal, currencyCode, maxDigits),
+                        Res.string.feature_loan_general_detail_disbursed_amount to CurrencyFormatter.format(loan.principal, currencyCode, maxDigits),
+                    ),
+                ),
+                summaryRows = listOf(
+                    LoanAccountGeneralState.SummaryRowState(
+                        component = getString(Res.string.feature_loan_general_summary_row_principal),
+                        original = CurrencyFormatter.format(summary?.principalDisbursed, currencyCode, maxDigits),
+                        paid = CurrencyFormatter.format(summary?.principalPaid, currencyCode, maxDigits),
+                        waived = CurrencyFormatter.format(summary?.principalWaived, currencyCode, maxDigits),
```
biplab1: This is repetitive, introduce a helper function:
```Kotlin
fun format(value: Number?) = CurrencyFormatter.format(value, currencyCode, maxDigits)
```

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/LoanAccountGeneralRepositoryImp.kt
biplab1, line 1: This implementation mixes concerns (network state handling, error mapping, and data fetching) and may lead to incorrect behavior (e.g., `combine` triggering network calls even when offline). It would be better to align on the expected architecture and responsibilities first.

I’d suggest discussing this approach with @niyajali on Slack before proceeding with further implementation.

biplab1:

biplab1:
@kartikey004 @amanna13 Please upload the latest screen recording.

core/data/src/commonMain/kotlin/com/mifos/core/data/mappers/loan/LoanAccountGeneralMapper.kt
```diff
@@ -0,0 +1,193 @@
...
+        name = name,
+        decimalPlaces = decimalPlaces,
+        inMultiplesOf = inMultiplesOf,
+        displaySymbol = displaySymbol,
+        nameCode = nameCode,
+        displayLabel = displayLabel,
+    )
+}
+
+private fun LoanTimelineEntity.toModel(): LoanTimeline {
```
biplab1: Please follow the naming conventions. Here, it should be `toDomain`. Also, check the naming of the data classes.

core/data/src/commonMain/kotlin/com/mifos/core/data/mappers/loan/LoanAccountGeneralMapper.kt
```diff
@@ -0,0 +1,193 @@
...
+private fun ActualDisbursementDateEntity.toModel(): ActualDisbursementDate {
+    return ActualDisbursementDate(
+        loanId = loanId,
+        year = year,
+        month = month,
+        date = date,
+    )
+}
+
+private fun LoansAccountSummaryEntity.toModel(): LoansAccountSummary {
```
biplab1: Same here, please review.

kartikeyshukla15:
@amanna13  can you please update the screen recording after fetching the latest changes on the branch as I think I can't do so .

kartikeyshukla15:
Updated recording - 
<video src="https://github.com/user-attachments/assets/d0770f03-972e-4c4e-8fc7-dca8a12e106e" controls height="200"></video>

kartikeyshukla15:
@biplab1 please tell if you also need any other changes. Also this one needs to get merged as two of my PRs are dependent on it.

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/LoanAccountGeneralRepositoryImp.kt
```diff
@@ -0,0 +1,38 @@
...
+class LoanAccountGeneralRepositoryImp(
+    private val dataManagerLoan: DataManagerLoan,
+    private val networkMonitor: NetworkMonitor,
+    private val ioDispatcher: CoroutineDispatcher,
+) : LoanAccountGeneralRepository {
+
+    override fun getLoanById(loanId: Int): Flow<DataState<LoanWithAssociations?>> {
+        return networkMonitor.withNetworkCheck(
+            dataManagerLoan.getLoanById(loanId)
+                .map { loan -> loan?.let(LoanAccountGeneralMapper::mapFromEntity) }
```
biplab1: Instead of mapping from Entity, it should map from DTO. Please review. DTO should be used in the network layer.

core/model/src/commonMain/kotlin/com/mifos/core/model/entity/loan/loanWithAssociations/ActualDisbursementDate.kt
```diff
@@ -0,0 +1,23 @@
...
+import kotlinx.serialization.Serializable
+
+@Parcelize
+@Serializable
+data class ActualDisbursementDate(
+    val loanId: Int? = null,
+    val year: Int? = null,
+    val month: Int? = null,
+    val date: Int? = null,
+) : Parcelable
```
biplab1: Domain models do not require `@Parcelize` or `@Serializable` here. Please review.

core/model/src/commonMain/kotlin/com/mifos/core/model/entity/loan/loanWithAssociations/LoanAccountSummary.kt
```diff
@@ -0,0 +1,55 @@
...
+ * See https://github.com/openMF/mifos-x-field-officer-app/blob/master/LICENSE.md
+ */
+package com.mifos.core.model.entity.loan.loanWithAssociations
+
+import com.mifos.core.model.utils.Parcelable
+import com.mifos.core.model.utils.Parcelize
+import kotlinx.serialization.Serializable
+
+@Parcelize
+@Serializable
```
biplab1: Same here, please review.

biplab1 (dismissed):
Please check all the data classes, and introduce DTOs, Entities, Domain models and Mappers appropriately. If service layer requires an `HttpResponse`, please add that. Map it to DTO in the data manager. Only DTOs should be annotated with `@Serializable`. No data classes should be annotated with `@Parcelize`.

biplab1:
@Kartikey15dem I don't think this is the correct way to handle the requested changes. 

This PR shouldn't contain changes from other modules. We should preserve the older data classes for modules which are outside the scope of this PR, and introduce new data classes as requested for this module only. Later, other modules will be refactored in separate PRs. Please review.

kartikeyshukla15:
> @Kartikey15dem I don't think this is the correct way to handle the requested changes.
> 
> This PR shouldn't contain changes from other modules. We should preserve the older data classes for modules which are outside the scope of this PR, and introduce new data classes as requested for this module only. Later, other modules will be refactored in separate PRs. Please review.

@biplab1 These changes came as a result of refactoring LoanWithAssociationsDto. That refactor introduced a series of dependent changes which were required to keep the app running correctly.

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,438 @@
...
+                    )
+                }
+            },
+            widths = colWidths,
+            backgroundColor = lerp(
+                KptTheme.colorScheme.surface,
+                KptTheme.colorScheme.primary,
+                0.3f,
+            ),
+            edgeOffset = DesignToken.padding.medium,
```
sahilshivekar: set this to KptTheme.spacing.md so that the other sections and the table will have the same padding.
It's looking odd that the other sections are having more padding than the table

Image

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,438 @@
...
+                                style = KptTheme.typography.bodySmall,
+                                fontWeight = FontWeight.Normal,
+                                textColor = textColor,
+                            )
+                        }
+                    }
+                },
+                widths = colWidths,
+                backgroundColor = KptTheme.colorScheme.surface,
+                edgeOffset = DesignToken.padding.medium,
```
sahilshivekar: set this to KptTheme.spacing.md here too

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanAccountGeneral/LoanAccountGeneralScreen.kt
```diff
@@ -0,0 +1,438 @@
...
+                            style = KptTheme.typography.bodyMedium,
+                            fontWeight = FontWeight.Bold,
+                            textColor = textColor,
+                        )
+                    }
+                }
+            },
+            widths = colWidths,
+            backgroundColor = lerp(KptTheme.colorScheme.surface, KptTheme.colorScheme.primary, 0.15f),
+            edgeOffset = DesignToken.padding.medium,
```
sahilshivekar: set this to KptTheme.spacing.md here too

sahilshivekar (changes requested):

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/LoanAccountGeneralRepositoryImp.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flowOn
+
+class LoanAccountGeneralRepositoryImp(
+    private val dataManagerLoan: DataManagerLoan,
+    private val networkMonitor: NetworkMonitor,
+    private val ioDispatcher: CoroutineDispatcher,
+) : LoanAccountGeneralRepository {
+
+    override fun getLoanById(loanId: Int): Flow<DataState<LoanWithAssociations?>> {
```
sahilshivekar: Don't use Flow here
Just keep it as DataState
Make use of runAsDataState as all the newly created repositoryImpl in the application are using it

refer to [LoanChargeOffRepositoryImpl.kt](https://github.com/sahilshivekar/mifos-x-field-officer-app/blob/development/core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/loan/LoanChargeOffRepositoryImpl.kt)

sahilshivekar (changes requested):

sahilshivekar:
Rest LGTM. I will approve once the requested change is implemented.

sahilshivekar (approved):

kartikeyshukla15:
<video src="https://github.com/user-attachments/assets/c931752a-fc9d-4d01-ae1f-a1a21b9d8d24" controls height="200"></video>

sahilshivekar (commented):
@revanthkumarJ I think this can be merged.

revanthkumarJ (approved):

PR #763, opened 2017-08-15
Fix: allow client name less than 4 characters

mayank-kgp:
Resolves issue #746

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Apply the `MifosStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #2659, opened 2026-03-31
feat(loan): implement disburse screen for approved loan account

kartikey004:
Fixes - [Jira-#681](https://mifosforge.jira.com/browse/MIFOSAC-681)

Note - Navigation to the General tab screen upon successful loan disbursement will be updated through [Jira-#763](https://mifosforge.jira.com/browse/MIFOSAC-763).

[Screen_recording_20260402_173058.webm](https://github.com/user-attachments/assets/4b8b2bd7-c878-4b2c-af49-0b83fdefe261)

## Summary by CodeRabbit

* **New Features**
  * Loan disbursement flow with a "Disburse Loan" action and new "Approved" profile status
  * Optional "Show payment details" toggle revealing multiple payment fields
  * Direct navigation between loan profile and disbursement screens

* **Improvements**
  * Shorter disbursement labels and currency-prefixed amount input
  * Network-aware template loading, dialog-based loading/error UX, and clearer submission feedback
  * New strings for available disbursement amount, unknown/invalid errors, and payment-field labels
* **Style**
  * New color for approved loan status in the theme


feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanDisbursement/LoanAccountDisbursementScreen.kt
```diff
@@ -11,334 +11,448 @@ package com.mifos.feature.loan.loanDisbursement
...
+        state.amount.toDoubleOrNull() != null
+
     val datePickerState = rememberDatePickerState(
-        initialSelectedDateMillis = disbursementDate,
+        initialSelectedDateMillis = state.disbursementDate,
         selectableDates = object : SelectableDates {
             override fun isSelectableDate(utcTimeMillis: Long): Boolean {
-                return utcTimeMillis >= Clock.System.now().toEpochMilliseconds()
+                return utcTimeMillis <= Clock.System.now().toEpochMilliseconds()
             }
```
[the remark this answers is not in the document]
kartikey004: similar pattern is implemented in `CreateNewClientScreen.kt`

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanDisbursement/LoanAccountDisbursementScreen.kt
```diff
@@ -11,334 +11,454 @@ package com.mifos.feature.loan.loanDisbursement
...
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+private fun LoanAccountDisbursementContent(
+    state: LoanDisbursementState,
+    onAction: (LoanDisbursementAction) -> Unit,
+    onCancel: () -> Unit,
+) {
+    val scrollState = rememberScrollState()
+    var showDatePickerDialog by rememberSaveable { mutableStateOf(false) }
```
itsPronay: Put that in viewmodel's ui state

itsPronay (changes requested):

itsPronay (changes requested):
image

why are we using `#` in the text field label? You can use the title "Bank Number" instead.
if these are numeric only text fields, make sure to set the appropriate keyboardOptions (example : number input only)

I remember @biplab1 suggested a better approach for handling this when reviewing someone else's PR, but I cant recall it at the moment.

kartikey004:
> why are we using `#` in the text field label? You can use the title "Bank Number" instead. if these are numeric only text fields, make sure to set the appropriate keyboardOptions (example : number input only)
> 
> I remember @biplab1 suggested a better approach for handling this when reviewing someone else's PR, but I cant recall it at the moment.

I wanted to keep it similar to web app hence used `#` 

image

itsPronay:
> I wanted to keep it similar to web app hence used `#`

yes, I understand. but correct me if i am wrong, In a web app, you cant always strictly prevent users from entering text into a number field (depending on the browser)

On mobile, however, its different. you can change the keyboard options to force a numeric keypad.

I think using labels like `Account Number`, `Bank Number`, or `Cheque Number` would be better and we should restrict the input to numbers only. (instead of using the # symbol inside the label).

kartikey004:
> > I wanted to keep it similar to web app hence used `#`
> 
> yes, I understand. but correct me if i am wrong, In a web app, you cant always strictly prevent users from entering text into a number field (depending on the browser)
> 
> On mobile, however, its different. you can change the keyboard options to force a numeric keypad.
> 
> I think using labels like `Account Number`, `Bank Number`, or `Cheque Number` would be better and we should restrict the input to numbers only. (instead of using the # symbol inside the label).

Yes, I understand. I will make the changes.

biplab1:
Is the following toggle switch consistent with the other feature screens?

image

biplab1:
After pressing the submit button on disburse screen, it does not show a proper feedback. Is it navigating to the right screen? Please review the web-app.

kartikey004:
> After pressing the submit button on disburse screen, it does not show a proper feedback. Is it navigating to the right screen? Please review the web-app.

As discussed in the standup last week, the navigation is currently set to the LoanAccountProfileScreen since the General tab screen PR has not been merged yet. I will create a separate ticket to address the navigation of the disbursement screen.

biplab1:
> > After pressing the submit button on disburse screen, it does not show a proper feedback. Is it navigating to the right screen? Please review the web-app.
> 
> As discussed in the standup last week, the navigation is currently set to the LoanAccountProfileScreen since the General tab screen PR has not been merged yet. I will create a separate ticket to address the navigation of the disbursement screen.

Please create a ticket and reference it here.

feature/loan/src/commonMain/composeResources/values/strings.xml
```diff
@@ -91,12 +91,22 @@
...
-    <string name="feature_loan_approval_disbursement_date">Approval Disbursement Date</string>
-    <string name="feature_loan_loan_amount_disbursed">Loan Amount Disbursed</string>
+    <string name="feature_loan_approval_disbursement_date">Disbursed On*</string>
+    <string name="feature_loan_loan_amount_disbursed">Transaction Amount*</string>
+    <string name="feature_loan_submission_failed">Failed to disburse loan. Please try again.</string>
     <string name="feature_loan_payment_type">Payment Type</string>
-    <string name="feature_loan_disbursement_note">Disbursement Note</string>
+    <string name="feature_loan_disbursement_note">Note</string>
     <string name="feature_loan_error_network_not_available">Network not available</string>
     <string name="feature_loan_error_amount_can_not_be_empty">Amount can not be empty</string>
```
biplab1:
```suggestion
    <string name="feature_loan_error_amount_cannot_be_empty">Amount cannot be empty</string>
```

feature/loan/src/commonMain/composeResources/values/strings.xml
```diff
@@ -91,12 +91,22 @@
...
+    <string name="feature_loan_available_disbursement_amount">Available Disbursement Amount (with Over Applied)</string>
+    <string name="feature_loan_show_payment_details">Show Payment Details</string>
+    <string name="feature_loan_show_account_number">Account Number</string>
+    <string name="feature_loan_cheque_number">Cheque Number</string>
+    <string name="feature_loan_routing_code">Routing Code</string>
+    <string name="feature_loan_receipt_number">Receipt Number</string>
+    <string name="feature_loan_bank_number">Bank Number</string>
+    <string name="feature_loan_unknown_error">Unknown Error</string>
```
biplab1: We should get rid of this unknown error string resource. I'm not sure why was this added in the first place. A specific error should be shown instead.

biplab1:
> Is the following toggle switch consistent with the other feature screens?
> image

@kartikey004 Can you please refer `mifos-mobile` feature modules and its design files to make UI components consistent across all projects?

feature/loan/src/commonMain/kotlin/com/mifos/feature/loan/loanDisbursement/LoanAccountDisbursementViewModel.kt
```diff
@@ -9,68 +9,278 @@
...
+    private var loadJob: Job? = null
 
-    val loanAccountDisbursementUiState: StateFlow<LoanAccountDisbursementUiState>
-        get() = _loanAccountDisbursementUiState
+    init {
+        observeNetworkAndLoad()
+    }
 
-    fun loadLoanTemplate() {
+    private fun observeNetworkAndLoad() {
```
biplab1: Can you tell us the reason for combining both observeNetwork and loadLoanTemplate? Currently, the naming doesn't capture what is being loaded.

kartikey004: to apply auto retry if the user network is bad. 
i will rename: `observeNetworkAndLoad()` to `observeNetworkAndLoadTemplate()` and 
`loadLoanTemplate `to `loadDisbursementLoanTemplate`

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/LoanAccountDisbursementRepositoryImp.kt
biplab1, line 1: This follows the approach suggested by @niyajali, but the current `combine` usage doesn’t actually prevent upstream calls when offline and may lead to incorrect error mapping.

I would suggest that we should align on the exact expected behavior/implementation details on Slack before proceeding, to avoid unintended side effects.

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/LoanAccountDisbursementRepositoryImp.kt
```diff
@@ -12,32 +12,65 @@ package com.mifos.core.data.repositoryImp
...
+    ): DataState<GenericResponse> {
+        return try {
+            if (!networkMonitor.isOnline.first()) {
+                return DataState.Error(NetworkUnavailableException())
+            }
+
+            val response = dataManagerLoan.disburseLoan(loanId, loanDisbursement)
+                .asDataStateFlow()
+                .first { it !is DataState.Loading }
+
```
biplab1, line 64: You're converting a one-shot API call into a Flow.
Then converting it back into a single value.

If API is one-shot -> keep it simple:
```Kotlin
return try {
    val response = dataManagerLoan.disburseLoan(...)
    DataState.Success(response)
} catch (e: Exception) {
    DataState.Error(e)
}
```
Or expose suspend -> not Flow

biplab1:
@coderabbitai full review

biplab1:
@kartikey004 Please fix the merge conflicts.

revanthkumarJ:
closed because this was implemented in https://github.com/openMF/mifos-x-field-officer-app/pull/2700

PR #2666, opened 2026-04-12
refactor(note): add DataState, optimize Flow, add DTOs, entities, domain models

itsPronay:
JIRA - https://mifosforge.jira.com/browse/MIFOSAC-769 

Screen Recording uploaded to Slack by @Arinyadav1:

https://github.com/user-attachments/assets/f171ca60-9bb8-4047-a84d-161968624805

## Summary by CodeRabbit

* **New Features**
  * Added explicit offline detection with a dedicated "network unavailable" error and consistent success/error state wrapping for async operations.
* **Bug Fixes**
  * Improved sync flow error handling to show a distinct offline message when there’s no connectivity and to reliably surface success or failure states to the UI.


biplab1:

core/data/src/commonMain/kotlin/com/mifos/core/data/util/NetworkUnavailableException.kt
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * See https://github.com/openMF/mifos-x-field-officer-app/blob/master/LICENSE.md
+ */
+package com.mifos.core.data.util
+
+/**
+ * Exception thrown when network is not available and the operation requires network connectivity.
+ */
+class NetworkUnavailableException(
+    message: String = "Network is unavailable. Please check your connection and try again.",
```
biplab1: Shouldn't this be using a string resource?

itsPronay: Yeah, I thought about that, but I don’t think we can extract strings from stringResource here without context. 

And we can't use `stringResource` directly as `message` only takes String type data
```kotlin
public expect open class Throwable {
    public open val message: String?
    public open val cause: Throwable?

    public constructor()
    public constructor(message: String?)
    public constructor(cause: Throwable?)
    public constructor(message: String?, cause: Throwable?)
}
```

biplab1: Have you tried using `getString(message: StringResource)` and passing the string value?

core/data/src/commonMain/kotlin/com/mifos/core/data/util/DataStateUtils.kt
biplab1, line 1: Please replace `asDataState` in comments with `runAsDataState`.

core/data/src/commonMain/kotlin/com/mifos/core/data/util/NetworkUnavailableException.kt
itsPronay: `getString()` comes under `compose.components.resources`, and our module `core.data` is purely based on kotlin, not Compose

biplab1: Yes, you are right.

biplab1:
@itsPronay Please upload screen recording of the feature screen - with and without network.

biplab1:
@Arinyadav1 @itsPronay Could you please address the @coderabbitai suggestions relevant to this PR?

feature/offline/src/commonMain/kotlin/com/mifos/feature/offline/syncGroupPayloads/SyncGroupPayloadsViewModel.kt
```diff
@@ -120,13 +123,24 @@ class SyncGroupPayloadsViewModel(
...
                 SyncGroupPayloadsUiState.Loading
-            try {
-                repository.createGroup(groupPayload!!)
-                deleteAndUpdateGroupPayload()
-            } catch (e: Exception) {
-                _syncGroupPayloadsUiState.value =
-                    SyncGroupPayloadsUiState.Error(Res.string.feature_offline_error_group_sync_failed)
-                updateGroupPayload()
+
+            when (val result = repository.createGroup(groupPayload!!)) {
```
biplab1: Please avoid using not-null assertion operator.

biplab1:
image

Shouldn't the screen stay where it is when clicking the submit button, after typing the note, since there is a "no internet" which is shown below instead of showing the try again page? Showing a try again page indicates the note being typed was lost, however, it is not. When the network is restored it has the note, which was just added, in the background. The UX could be improved a bit. 

This could be handled in a separate PR, I think.

core/data/src/commonMain/kotlin/com/mifos/core/data/util/NetworkUnavailableException.kt
```diff
@@ -0,0 +1,17 @@
...
+) : Throwable(message)
```
biplab1: Shouldn't this be extending `Exception` instead? The codebase has most of the exception handlers catching `Exception`, not `Throwable`.

```Kotlin
class NetworkUnavailableException(
    message: String = "Network is unavailable...",
) :  Exception(message)
```

biplab1:
Could you please take a look at this issue: https://github.com/openMF/mifos-x-field-officer-app/pull/2666#discussion_r3075012285

core/data/src/commonMain/kotlin/com/mifos/core/data/util/NetworkUnavailableException.kt
Arinyadav1: The DataState take throwable that's why here use throwable

biplab1: Okay, got it.

biplab1 (approved):

biplab1:
@Arinyadav1 Have you tested the relevant screens to verify if everything works fine after the recent updates?

biplab1:
@niyajali If everything looks good, please feel free to go ahead and merge this.

core/data/src/commonMain/kotlin/com/mifos/core/data/util/NetworkUnavailableException.kt
biplab1, line 1: Please compare this with that present in NetworkMonitor file shared by @niyajali to make sure all things are consistent.

biplab1 (changes requested):

core/data/src/commonMain/kotlin/com/mifos/core/data/util/NetworkUnavailableException.kt
biplab1: Please use this instead, taken from NetworkMonitory file https://mifos.slack.com/archives/C06FC73DZPW/p1777383001891109?thread_ts=1775837735.406679&cid=C06FC73DZPW:
```Kotlin
/**
 * Thrown when a network operation is attempted without a stable, validated
 * internet connection.
 *
 * @param message Human-readable reason shown in logs and crash reports.
 * @param cause   Optional underlying throwable.
 */
class NetworkUnavailableException(
    message: String = "No stable network connection available",
    cause: Throwable? = null,
) : IllegalStateException(message, cause)
```

biplab1:
@Arinyadav1 Have you tested the relevant screens to verify if everything works fine after the recent updates?

Arinyadav1:
@biplab1 Yes, It is working fine after changes 
image

core/data/src/androidMain/kotlin/com/mifos/core/data/util/ConnectivityManagerNetworkMonitor.kt
```diff
@@ -72,5 +72,5 @@ internal class ConnectivityManagerNetworkMonitor(
 
     private fun ConnectivityManager.isCurrentlyConnected() = activeNetwork
         ?.let(::getNetworkCapabilities)
-        ?.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) ?: false
+        ?.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED) ?: false
```
biplab1, line 75: I noticed it yesterday also but didn't think of it much. Could you please tell us the reason for this change? The `mifos-pay` project still uses `NET_CAPABILITY_INTERNET`. 

Ref: https://github.com/openMF/mifos-pay/blob/development/core/data/src/androidMain/kotlin/org/mifospay/core/data/util/ConnectivityManagerNetworkMonitor.kt

Arinyadav1: `NET_CAPABILITY_INTERNET` only indicates that a network (Wi-Fi or mobile data) is available and claims to provide internet access, but it does not guarantee actual connectivity.

For example, a device can be connected to Wi-Fi without internet access, and this capability would still return true.

To ensure real internet connectivity, we should also check `NET_CAPABILITY_VALIDATED`, which confirms that the system has successfully verified access to the internet.

biplab1 (dismissed):

biplab1:
@niyajali If everything looks good, please feel free to go ahead and merge this.

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/NoteRepositoryImp.kt
```diff
@@ -12,53 +12,77 @@ package com.mifos.core.data.repositoryImp
...
+    private val dispatcher: DispatcherManager,
 ) : NoteRepository {
 
     override suspend fun addNewNote(
         resourceType: String,
         resourceId: Long,
         notesPayload: NotesPayload,
-    ): GenericResponse {
-        return dataManagerNote.addNewNote(resourceType, resourceId, notesPayload)
+    ): DataState<GenericResponse> {
```
niyajali: Create DTO, Entity, Domain Model class and appropriate mapper and return accordingly on each service/repository/API interfaces..

core/data/src/commonMain/kotlin/com/mifos/core/data/util/DataStateUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+    block: suspend () -> T,
+): DataState<T> =
+    try {
+        if (context != null) {
+            DataState.Success(withContext(context) { block() })
+        } else {
+            DataState.Success(block())
+        }
+    } catch (e: Throwable) {
+        if (e is CancellationException) throw e
```
niyajali: No need to throw this.. remove it for now later we'll find the proper approach

feature/note/src/commonMain/kotlin/com/mifos/feature/note/addEditNotes/AddEditNoteViewModel.kt
```diff
@@ -137,48 +83,75 @@ class AddEditNoteViewModel(
...
-                                    )
-                                }
-                            }
-
-                            DataState.Loading -> {
-                                mutableStateFlow.update {
-                                    it.copy(dialogState = AddEditNoteState.DialogState.Loading)
-                                }
-                            }
+                when (val dataState = updateNoteUseCase(type, route.resourceId.toLong(), id, notesPayload)) {
```
niyajali: Create an internal sealed interface and action and send the action and manage it properly check out the wallet project.. properly read some ViewModels and other files..

feature/offline/src/commonMain/kotlin/com/mifos/feature/offline/syncGroupPayloads/SyncGroupPayloadsViewModel.kt
```diff
@@ -120,13 +123,27 @@ class SyncGroupPayloadsViewModel(
...
-            try {
-                repository.createGroup(groupPayload!!)
-                deleteAndUpdateGroupPayload()
-            } catch (e: Exception) {
-                _syncGroupPayloadsUiState.value =
-                    SyncGroupPayloadsUiState.Error(Res.string.feature_offline_error_group_sync_failed)
-                updateGroupPayload()
+
+            groupPayload?.let { payload ->
+                when (val result = repository.createGroup(payload)) {
```
niyajali: Similarly here send this result to an internal action and manage it properly so it will become clean and properly readable and maintainable

niyajali (dismissed):
There are few more enhancement needed.. please check out the wallet project and update accordingly..

feature/offline/src/commonMain/kotlin/com/mifos/feature/offline/syncGroupPayloads/SyncGroupPayloadsViewModel.kt
Arinyadav1: This file use previous `ViewModel()` approach that's why it will fix in separate PR and the whole new implementation approach is implemented in `note` module

biplab1: We should either migrate this feature to MVI or remove it completely since other contributors will refer to this PR for their own PRs. There should be no ambiguity about the app architecture pattern.

core/network/src/commonMain/kotlin/com/mifos/core/network/services/NoteService.kt
```diff
@@ -32,14 +31,14 @@ interface NoteService {
...
-     * @return [GenericResponse]
+     * @return [Unit]
      */
     @POST("{resourceType}/{resourceId}/" + APIEndPoint.NOTES)
     suspend fun addNewNote(
         @Path("resourceType") resourceType: String,
         @Path("resourceId") resourceId: Long,
         @Body notesPayload: NotesPayload,
-    ): GenericResponse
+    ): Unit
```
biplab1: Could you please tell us the reason for replacing `GenericResponse` with `Unit`?

core/network/src/commonMain/kotlin/com/mifos/core/network/datamanager/DataManagerNote.kt
```diff
@@ -22,7 +21,7 @@ class DataManagerNote(
         resourceType: String,
         resourceId: Long,
         notesPayload: NotesPayload,
-    ): GenericResponse {
+    ) {
```
biplab1: Same here, please clarify.

core/network/src/commonMain/kotlin/com/mifos/core/network/services/NoteService.kt
Arinyadav1: Because the one shot api doesn't return anything relevant data so instead of use `GenericResponse` I have changed it to `Unit`

core/network/build.gradle.kts
```diff
@@ -41,7 +41,7 @@ kotlin{
             implementation(libs.ktor.client.auth)
             implementation(libs.ktor.serialization.kotlinx.json)
             implementation(libs.ktorfit.lib)
-            implementation(libs.squareup.okio)
+             implementation(libs.squareup.okio)
```
biplab1: Please fix the indentation.

feature/note/src/commonMain/kotlin/com/mifos/feature/note/addEditNotes/AddEditNoteViewModel.kt
```diff
@@ -282,4 +271,13 @@ sealed interface AddEditNoteAction {
     data object DismissDialog : AddEditNoteAction
     data object MisTouchBackDialog : AddEditNoteAction
```
biplab1: Could you please improve the naming of data objects: `MisTouchBack` and `MisTouchBackDIalog`?

core/network/src/commonMain/kotlin/com/mifos/core/network/datamanager/DataManagerNote.kt
Arinyadav1: Because the one shot api doesn't return anything relevant data so instead of use `GenericResponse `I have changed it to `Unit`

feature/offline/src/commonMain/kotlin/com/mifos/feature/offline/syncGroupPayloads/SyncGroupPayloadsViewModel.kt
Arinyadav1: I will revert it back and only note module have latest implementation

Arinyadav1:
@biplab1 please do the final review

core/network/src/commonMain/kotlin/com/mifos/core/network/services/NoteService.kt
biplab1: As discussed in the standup please revert the change. We should keep `GenericResponse` since the server does send some responses, even though we don't use it now, but could be used later.

Arinyadav1: okay

Arinyadav1: done review please

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/NoteRepositoryImp.kt
biplab1: @Arinyadav1 Looks like this was not taken care of. Please review.

core/network/src/commonMain/kotlin/com/mifos/core/network/services/NoteService.kt
```diff
@@ -100,7 +100,7 @@ interface NoteService {
      * @param resourceType resourceType, eg : Client, Loan, Group, Savings Account
      * @param resourceId resourceId, eg : ClientId, LoanId, GroupId, Savings AccountId
      * @param noteId noteId
-     * @param notesPayload
+     * @param Unit
```
biplab1: The changes in the comments needs to be reverted as well. Please check other places in this file.

biplab1: It is better to do a self-review after pushing changes to the PR.

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/NoteRepositoryImp.kt
biplab1: Instead of `GenericResponse`, please create dedicated data classes as mentioned above to handle responses related to Notes. 

Ref: https://fineract.apache.org/docs/legacy/index.html, search "Notes"

biplab1: Adding them here for quick reference:

Screenshot 2026-04-30 093609
Screenshot 2026-04-30 093623
Screenshot 2026-04-30 093635
Screenshot 2026-04-30 093643
Screenshot 2026-04-30 093651
Screenshot 2026-04-30 093659

Arinyadav1:
@biplab1 please do the final review

biplab1:
@Arinyadav1 I can see you have added one data class `NoteResponse` and I cannot find the rest of the requested data classes. Have they already been implemented? I also do not see any mapper either.

It is much more convenient for the reviewer if things are explicitly mentioned to aid the review process. Otherwise, we need to go look for things in the codebase which the contributor might have already observed during their own research and while implementing the code.

biplab1:
Data classes types:
- API -> DTO (network model)
- DB -> Entity (database model)
- Business Logic/ UI -> Domain model (business or app model), in our codebase it is simply called model.

Data classes' locations:
- DTOs live in the `:core:network:dto`,
- Domain models (aka models) live inside `:core:model`,
- Mappers live inside the `:core:data:mappers` and
- Entities live in `:core:database:objects`.

Types of Mappers: DTO --> Domain Models, and Entity <--> Domain Models

> **_NOTE:_** This codebase does not have proper separation of concerns for the data classes. An attempt is being made to improve that.

## Here is an example:
 (PLEASE DO NO USE IT AS IT IS, VERIFY AGAINST THE APIs AND RESPONSES)
 
## 1. Domain model (core/model)
Main model used everywhere in app/UI

```kotlin
data class Note(
    val id: Int,
    val resourceId: Int,
    val resourceType: NoteResourceType,

    val note: String,
    val type: String,

    val createdBy: String,
    val createdOn: Long,

    val updatedBy: String?,
    val updatedOn: Long?
)

enum class NoteResourceType {
    CLIENT,
    LOAN,
    GROUP,
    SAVINGS
}
```

## 2. Network DTOs (core/network/dto)

### A. Note Response DTO

Used for:

* GET single note
* GET list of notes

```kotlin
@Serializable
data class NoteDto(
    val id: Int? = null,

    val clientId: Int? = null,
    val resourceId: Int? = null,

    val noteType: NoteTypeDto? = null,
    val note: String? = null,

    val createdById: Int? = null,
    val createdByUsername: String? = null,
    val createdOn: Long? = null,

    val updatedById: Int? = null,
    val updatedByUsername: String? = null,
    val updatedOn: Long? = null
)
```

### B. Nested Type DTO

```kotlin
@Serializable
data class NoteTypeDto(
    val id: Int? = null,
    val code: String? = null,
    val value: String? = null
)
```

### C. Create / Update Request DTO

Used for:

* POST
* PUT (UPDATE)

```kotlin
@Serializable
data class NoteRequestDto(
    val note: String
)
```

### D. Create Note Response DTO

```kotlin
@Serializable
data class CreateNoteResponseDto(
    val officeId: Int? = null,
    val clientId: Int? = null,
    val resourceId: Int? = null
)
```

### E. Update Note Response DTO

```kotlin
@Serializable
data class UpdateNoteResponseDto(
    val officeId: Int? = null,
    val clientId: Int? = null,
    val resourceId: Int? = null,
    val changes: ChangesDto? = null
)

@Serializable
data class ChangesDto(
    val note: String? = null
)
```

### F. Delete Note Response DTO

```kotlin
@Serializable
data class DeleteNoteResponseDto(
    val resourceId: Int? = null
)
```

## 3. Database Entity (core/database/entity)

Room version

```kotlin
@Entity(tableName = "notes")
data class NoteEntity(
    @PrimaryKey val id: Int,

    val resourceId: Int,
    val resourceType: String,

    val note: String,
    val type: String,

    val createdBy: String,
    val createdOn: Long,

    val updatedBy: String?,
    val updatedOn: Long?
)
```

## 4. Mappers (core/data/mappers)

### DTO → Domain Model 
- Name it as `NoteDtoMapper.kt`

```kotlin
fun NoteDto.toDomain(resourceType: NoteResourceType): Note {
    return Note(
        id = id ?: 0,
        resourceId = resourceId ?: clientId ?: 0,
        resourceType = resourceType,

        note = note.orEmpty(),
        type = noteType?.value.orEmpty(),

        createdBy = createdByUsername.orEmpty(),
        createdOn = createdOn ?: 0L,

        updatedBy = updatedByUsername,
        updatedOn = updatedOn
    )
}
```

### Domain Model → Entity 
- Put it inside `NoteEntityMapper.kt`
```kotlin
fun Note.toEntity(): NoteEntity {
    return NoteEntity(
        id = id,
        resourceId = resourceId,
        resourceType = resourceType.name,

        note = note,
        type = type,

        createdBy = createdBy,
        createdOn = createdOn,

        updatedBy = updatedBy,
        updatedOn = updatedOn
    )
}
```

### Entity → Domain Model
- Put it inside `NoteEntityMapper.kt`
```kotlin
fun NoteEntity.toDomain(): Note {
    return Note(
        id = id,
        resourceId = resourceId,
        resourceType = NoteResourceType.valueOf(resourceType),

        note = note,
        type = type,

        createdBy = createdBy,
        createdOn = createdOn,

        updatedBy = updatedBy,
        updatedOn = updatedOn
    )
}
```

## 5. Resource Wrapper 

(Optional but recommended) Since API uses `{resource}` dynamically:

```kotlin
enum class ResourcePath(val value: String) {
    CLIENT("clients"),
    LOAN("loans"),
    GROUP("groups"),
    SAVINGS("savingsaccounts")
}
```

Let me know if you have any questions.

biplab1:
Initially, I didn't think much about the payload, maybe I overlooked it completely.  You had mostly implemented it but let's handle it this way:

1. We have `feature:note`, so change `core:model:notes` to `core:model:note`, to stay consistent.
2. Use `CreateNoteInput` in Screen/ ViewModel and place it inside `core:model:note` - it is a Domain model and not annotated with `@Serializable`
3. Create `NoteRequestDto` in `core:network:dto`, which is annotated with `@Serializable`.
4.  In data layer we map `CreateNoteInput` to `NoteRequestDto` using a mapper `NoteInputMapper.kt`, which is placed inside `core:data:mappers`.
5.  Get rid of `NotesPayload`.

Let me know if you have any questions.

cmp-shared/cmp_shared.podspec
biplab1, line 1: Please revert this change.

core/database/src/commonMain/kotlin/com/mifos/room/entities/noncore/NoteEntity.kt
biplab1: I think we can remove `Parcelable` and `@Parcelize` since we won't be using Entity to transfer data anymore.

Ref: https://developer.android.com/guide/components/activities/parcelables-and-bundles

> [Parcelable](https://developer.android.com/reference/android/os/Parcelable) and [Bundle](https://developer.android.com/reference/android/os/Bundle) objects are intended to be used across process boundaries such as with IPC/Binder transactions, between activities with intents, and to store transient state across configuration changes.
>
> Note: [Parcel](https://developer.android.com/reference/android/os/Parcel) is not a general-purpose serialization mechanism, and you should never store any [Parcel](https://developer.android.com/reference/android/os/Parcel) data on disk or send it over the network.

cmp-shared/cmp_shared.podspec
Arinyadav1: it is automatically change while the synching of gradle

Arinyadav1: and without it throw macos build error

Arinyadav1:
@biplab1 do the final review

core/database/src/commonMain/kotlin/com/mifos/room/entities/noncore/NoteEntity.kt
Arinyadav1: done

cmp-shared/cmp_shared.podspec
biplab1: The MacOS desktop app build passed but workflow failed due to some Github's own internal issue, please revert this change. This file should not be a part of this PR.

And yes, this file gets auto-generated during a Gradle sync, but we need to revert the changes in this file before making a commit.

Ref: https://github.com/openMF/mifos-x-field-officer-app/actions/runs/25329796689/job/74260757049

Image

core/network/src/commonMain/kotlin/com/mifos/core/network/dto/note/UpdateNoteResponseDto.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+package com.mifos.core.network.dto.note
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class UpdateNoteResponseDto(
+    val officeId: Int? = null,
+    val clientId: Int? = null,
+    val resourceId: Int? = null,
```
biplab1: Please verify it these should be `Long` or `Int`.

core/network/src/commonMain/kotlin/com/mifos/core/network/dto/note/CreateNoteResponseDto.kt
```diff
@@ -0,0 +1,19 @@
...
+ */
+package com.mifos.core.network.dto.note
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class CreateNoteResponseDto(
+    val officeId: Int? = null,
+    val clientId: Int? = null,
+    val resourceId: Int? = null,
```
biplab1: Same here, please review.

core/network/src/commonMain/kotlin/com/mifos/core/network/dto/note/DeleteNoteResponseDto.kt
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * See https://github.com/openMF/mifos-x-field-officer-app/blob/master/LICENSE.md
+ */
+package com.mifos.core.network.dto.note
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class DeleteNoteResponseDto(
+    val resourceId: Int? = null,
```
biplab1: Same here, please review.

core/network/src/commonMain/kotlin/com/mifos/core/network/dto/note/UpdateNoteResponseDto.kt
Arinyadav1: It should be Long

Arinyadav1: Everywhere in project use `Long` for id

core/network/src/commonMain/kotlin/com/mifos/core/network/dto/note/NoteDto.kt
```diff
@@ -0,0 +1,49 @@
...
+ * @param note
+ * @param updatedById
+ * @param updatedByUsername
+ * @param updatedOn
+ */
+
+@Serializable
+data class NoteDto(
+
+    val clientId: Long? = null,
```
biplab1, line 33: Please check the API response. For example, `noteType` seems to be missing.

core/network/src/commonMain/kotlin/com/mifos/core/network/dto/note/CreateNoteResponseDto.kt
Arinyadav1: tell the reason above

core/network/src/commonMain/kotlin/com/mifos/core/network/dto/note/DeleteNoteResponseDto.kt
Arinyadav1: told the reason above

core/network/src/commonMain/kotlin/com/mifos/core/network/dto/note/NoteDto.kt
Arinyadav1: There is no use of `noteType` that's why I do not add in it

core/network/src/commonMain/kotlin/com/mifos/core/network/dto/note/UpdateNoteResponseDto.kt
biplab1: Could you please check what does web-app use?

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/NoteRepositoryImp.kt
```diff
@@ -11,54 +11,93 @@ package com.mifos.core.data.repositoryImp
...
+        ).flowOn(dispatcher.io)
 
     override suspend fun updateNote(
         resourceType: String,
         resourceId: Long,
         noteId: Long,
-        notesPayload: NotesPayload,
-    ): GenericResponse {
-        return dataManagerNote.updateNote(resourceType, resourceId, noteId, notesPayload)
+        createNoteInput: CreateNoteInput,
```
biplab1: Please create `UpdateNoteInput` and use it here. Also, add a corresponding mapper in `NoteInputMapper`. The content of the data classes for create and update are same and both should share the same `NoteRequestDto`, however, their use cases are different.

For example:
``` Kotlin
// Update
data class UpdateNoteInput(
    val note: String
)

// Mappers
fun CreateNoteInput.toDto(): NoteRequestDto
fun UpdateNoteInput.toDto(): NoteRequestDto
```

core/network/src/commonMain/kotlin/com/mifos/core/network/dto/note/NoteDto.kt
biplab1: We should consume everything that is available in the API response in the network layer using the DTO, and as per app requirement we can map it to the Domain model, where we can choose what to keep.

core/data/src/commonMain/kotlin/com/mifos/core/data/repositoryImp/NoteRepositoryImp.kt
Arinyadav1: done

core/network/src/commonMain/kotlin/com/mifos/core/network/dto/note/NoteDto.kt
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @param clientId
+ * @param createdById
+ * @param createdByUsername
+ * @param createdOn
+ * @param id
+ * @param note
+ * @param updatedById
+ * @param updatedByUsername
+ * @param updatedOn
```
biplab1: Please update parameters in the DocString also accordingly.

Arinyadav1: okay

biplab1:
@Arinyadav1 Have you tested the screens if they are working fine after the recent updates? Please upload a new screen recording.

biplab1 (approved):

PR #2679, opened 2026-05-10
Harden authentication flow with login throttling and safer credential handling

Divyateja2709:
hi @therajanmaurya 

This PR introduces a focused security remediation in the authentication flow, aligned with Cipher 2026 hardening goals.
It adds temporary login throttling after repeated failed attempts, avoids persisting plaintext passwords in local user state, and preserves password fidelity by removing accidental trimming before auth requests.

Changes included
Add failed-attempt lockout logic in login flow (5 failures -> 1 minute lockout).
Add localized error message for temporary lockout.
Stop storing raw password in persisted User object.
Keep username normalization (trim) but do not trim password in repository login call.
Why this matters
Reduces risk of brute-force behavior on client login attempts.
Minimizes sensitive data exposure in local persisted state.
Prevents authentication edge cases caused by password mutation.
Validation
Static lints on touched files: no issues.
Build verification was blocked locally because JAVA_HOME is not set in the environment.

### DIFF
--- a/feature/auth/src/commonMain/kotlin/com/mifos/feature/auth/login/LoginViewModel.kt
+++ b/feature/auth/src/commonMain/kotlin/com/mifos/feature/auth/login/LoginViewModel.kt
@@ -12,6 +12,7 @@ package com.mifos.feature.auth.login
 import androidclient.feature.auth.generated.resources.Res
 import androidclient.feature.auth.generated.resources.feature_auth_error_login_failed
 import androidclient.feature.auth.generated.resources.feature_auth_error_password_length
+import androidclient.feature.auth.generated.resources.feature_auth_error_too_many_attempts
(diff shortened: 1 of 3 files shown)


IOhacker (changes requested):
The Max failed Attempts is managed by the back end not by the front end. It should be removed.

feature/auth/src/commonMain/kotlin/com/mifos/feature/auth/login/LoginViewModel.kt
```diff
@@ -111,4 +127,21 @@ class LoginViewModel(
 
         _loginUiState.value = LoginUiState.PassCodeActivityIntent
     }
+
+    private fun isLoginTemporarilyLocked(): Boolean {
```
IOhacker, line 131: The Max failed Attempts is managed by the back end not by the front end. It should be removed.

PR #2704, opened 2026-08-21
kmp-project-template adoption

therajanmaurya:
Draft.


therajanmaurya:
Superseded by epic **`template-resync-from-dev`** — closing in favour of a resync branched from `dev`.

### Why

A `/kmp-project-template-sync` STEP 0.ANALYZE run (2026-09-19) measured path collision for both candidate bases against `template/dev`:

| base | collide | of which real | fork-only (survives) | template-only (arrives) |
|---|---:|---:|---:|---:|
| `dev` (`44d92532f`) | 134 | **95** (39 byte-identical) | 1,400 | 1,760 |
| this branch | **987** | 987 | 1,330 | 907 |

Starting from `dev` is roughly **10× less reconciliation work**. The cause is mechanical rather than a judgement about the work here: this branch *imported* the 2026-08-21 template trees (`core/`, `core-base/`, `scripts/`, `deployment/`) into the fork; those then aged five weeks, so every one of them is now a three-way merge. On `dev` most of those paths do not exist yet, so the same files simply arrive instead of conflicting.

### Root cause the analysis surfaced

The fork carries the **module-scoped** ownership contract (`core/** → owner: template`), which cannot distinguish the ~1,098 `com.mifos.*` domain files from the ~103 `kpt.core.*` template files sharing those Gradle modules — so every sync presented as an all-or-nothing diff. The current template replaced this with ~90 **package-scoped** rules plus a template-first default; `sync-ownership-transition` reports `loosening / fork-only = 1253`, i.e. the new contract correctly releasing fork code the old one wrongly claimed. Adopting that contract is the unlock, and it is one of five declared migrations in the template's own `template-migrations.yaml` — all five currently un-adopted here, two of them `blocking: true`.

### What replaces this

Epic `template-resync-from-dev` — 8 sequenced sub-plans, each gated on a detector verdict rather than on files touched:

1. baseline from `dev` · 2. `whitelabel-engine-layout` · 3. `module-deps-seam` · 4. `template-first-ownership` · 5. mechanical sync · 6. `ksp-registration` · 7. pure-sync reconciliation · 8. build green on android/ios/desktop/web

**This branch is retained, not deleted** — its history stays readable, including the unpushed `e60c74925` (`fix(deployment): retire *-upload-keystore-keypass`) should we want to recover it separately.

No reflection on the work in this PR; the base was simply the expensive one.

